PR #29, opened 2020-11-18
Enforce CSS byte count limit across all inline style additions and inline object-* attributes

### LINKED ISSUE #6: Copy object-position & object-fit into style attribute to reduce LCP
This is moved over from outstanding issues on the PR in the `ampproject/amp-wp repository` (https://github.com/ampproject/amp-wp/pull/5350#discussion_r508887832).

Via @westonruter:

> They are valid on `amp-img`. If supplied on an `img`, then the `AMP_Image_Sanitizer` passes them through.
>
> Nevertheless, for the purposes of SSR images, given this non-optimized input:
>
> ```html
> <amp-img src="https://amp.dev/static/img/icons/icon-512x512.png" width="512" height="250" object-fit="cover" object-position="right top"></amp-img>
> ```
>
> The Node.js Optimizer does this for SSR:
>
> ```html
> <amp-img src="https://amp.dev/static/img/icons/icon-512x512.png" width="512" height="250" object-fit="cover" object-position="right top" i-amphtml-ssr data-hero class="i-amphtml-layout-fixed i-amphtml-layout-size-defined" style="width:512px;height:250px;" i-amphtml-layout="fixed">
> 	
> </amp-img>
> ```
> 
> So it _is_ copying the `object-fit` and `object-position` attributes. Nevertheless, the Node.js Optimizer does not seem to be right here either. I can see that the AMP runtime is _also_ adding an inline `style` to the `img`: `object-fit: cover; object-position: right top;`. This is causing some LCP. So in reality, the `object-position` and `object-fit` should instead be copied into a `style` attribute like so:
> 
> ```html
> <amp-img src="https://amp.dev/static/img/icons/icon-512x512.png" width="512" height="250" object-fit="cover" object-position="right top" i-amphtml-ssr data-hero class="i-amphtml-layout-fixed i-amphtml-layout-size-defined" style="width:512px;height:250px;" i-amphtml-layout="fixed">
> 	
> </amp-img>
> ```
> 
> This will eliminate any negative LCP once the runtime initializes.
> 
> This also needs to be fixed in the Node.js Optimizer.
> 
> See [AMP Playground example](https://playground.amp.dev/#share=PCFkb2N0eXBlIGh0bWw+CjxodG1sIOKaoSBpLWFtcGh0bWwtbGF5b3V0IGktYW1waHRtbC1uby1ib2lsZXJwbGF0ZSB0cmFuc2Zvcm1lZD0ic2VsZjt2PTEiPgo8aGVhZD4KICA8bWV0YSBjaGFyc2V0PSJ1dGYtOCI+PHN0eWxlIGFtcC1ydW50aW1lIGktYW1waHRtbC12ZXJzaW9uPSIwMTIwMDkxMTIwMzQwMDIiPmh0bWx7b3ZlcmZsb3cteDpoaWRkZW4haW1wb3J0YW50fWh0bWwuaS1hbXBodG1sLWZpZXtoZWlnaHQ6MTAwJSFpbXBvcnRhbnQ7d2lkdGg6MTAwJSFpbXBvcnRhbnR9aHRtbDpub3QoW2FtcDRhZHNdKSxodG1sOm5vdChbYW1wNGFkc10pIGJvZHl7aGVpZ2h0OmF1dG8haW1wb3J0YW50fWh0bWw6bm90KFthbXA0YWRzXSkgYm9keXttYXJnaW46MCFpbXBvcnRhbnR9Ym9keXstd2Via2l0LXRleHQtc2l6ZS1hZGp1c3Q6MTAwJTstbW96LXRleHQtc2l6ZS1hZGp1c3Q6MTAwJTstbXMtdGV4dC1zaXplLWFkanVzdDoxMDAlO3RleHQtc2l6ZS1hZGp1c3Q6MTAwJX1odG1sLmktYW1waHRtbC1zaW5nbGVkb2MuaS1hbXBodG1sLWVtYmVkZGVkey1tcy10b3VjaC1hY3Rpb246cGFuLXk7dG91Y2gtYWN0aW9uOnBhbi15fWh0bWwuaS1hbXBodG1sLWZpZT5ib2R5LGh0bWwuaS1hbXBodG1sLXNpbmdsZWRvYz5ib2R5e292ZXJmbG93OnZpc2libGUhaW1wb3J0YW50fWh0bWwuaS1hbXBodG1sLWZpZTpub3QoLmktYW1waHRtbC1pbmFib3gpPmJvZHksaHRtbC5pLWFtcGh0bWwtc2luZ2xlZG9jOm5vdCguaS1hbXBodG1sLWluYWJveCk+Ym9keXtwb3NpdGlvbjpyZWxhdGl2ZSFpbXBvcnRhbnR9aHRtbC5pLWFtcGh0bWwtd2Vidmlldz5ib2R5e292ZXJmbG93LXg6aGlkZGVuIWltcG9ydGFudDtvdmVyZmxvdy15OnZpc2libGUhaW1wb3J0YW50O21pbi1oZWlnaHQ6MTAwdmghaW1wb3J0YW50fWh0bWwuaS1hbXBodG1sLWlvcy1lbWJlZC1sZWdhY3k+Ym9keXtvdmVyZmxvdy14OmhpZGRlbiFpbXBvcnRhbnQ7b3ZlcmZsb3cteTphdXRvIWltcG9ydGFudDtwb3NpdGlvbjphYnNvbHV0ZSFpbXBvcnRhbnR9aHRtbC5pLWFtcGh0bWwtaW9zLWVtYmVke292ZXJmbG93LXk6YXV0byFpbXBvcnRhbnQ7cG9zaXRpb246c3RhdGljfSNpLWFtcGh0bWwtd3JhcHBlcntvdmVyZmxvdy14OmhpZGRlbiFpbXBvcnRhbnQ7b3ZlcmZsb3cteTphdXRvIWltcG9ydGFudDtwb3NpdGlvbjphYnNvbHV0ZSFpbXBvcnRhbnQ7dG9wOjAhaW1wb3J0YW50O2xlZnQ6MCFpbXBvcnRhbnQ7cmlnaHQ6MCFpbXBvcnRhbnQ7Ym90dG9tOjAhaW1wb3J0YW50O21hcmdpbjowIWltcG9ydGFudDtkaXNwbGF5OmJsb2NrIWltcG9ydGFudH1odG1sLmktYW1waHRtbC1pb3MtZW1iZWQuaS1hbXBodG1sLWlvcy1vdmVyc2Nyb2xsLGh0bWwuaS1hbXBodG1sLWlvcy1lbWJlZC5pLWFtcGh0bWwtaW9zLW92ZXJzY3JvbGw+I2ktYW1waHRtbC13cmFwcGVyey13ZWJraXQtb3ZlcmZsb3ctc2Nyb2xsaW5nOnRvdWNoIWltcG9ydGFudH0jaS1hbXBodG1sLXdyYXBwZXI+Ym9keXtwb3NpdGlvbjpyZWxhdGl2ZSFpbXBvcnRhbnQ7Ym9yZGVyLXRvcDoxcHggc29saWQgdHJhbnNwYXJlbnQhaW1wb3J0YW50fSNpLWFtcGh0bWwtd3JhcHBlcitib2R5e3Zpc2liaWxpdHk6dmlzaWJsZX0jaS1hbXBodG1sLXdyYXBwZXIrYm9keSAuaS1hbXBodG1sLWxpZ2h0Ym94LWVsZW1lbnQsI2ktYW1waHRtbC13cmFwcGVyK2JvZHlbaS1hbXBodG1sLWxpZ2h0Ym94XXt2aXNpYmlsaXR5OmhpZGRlbn0jaS1hbXBodG1sLXdyYXBwZXIrYm9keVtpLWFtcGh0bWwtbGlnaHRib3hdIC5pLWFtcGh0bWwtbGlnaHRib3gtZWxlbWVudHt2aXNpYmlsaXR5OnZpc2libGV9I2ktYW1waHRtbC13cmFwcGVyLmktYW1waHRtbC1zY3JvbGwtZGlzYWJsZWQsLmktYW1waHRtbC1zY3JvbGwtZGlzYWJsZWR7b3ZlcmZsb3cteDpoaWRkZW4haW1wb3J0YW50O292ZXJmbG93LXk6aGlkZGVuIWltcG9ydGFudH1hbXAtaW5zdGFncmFte3BhZGRpbmc6NTRweCAwcHggMHB4IWltcG9ydGFudDtiYWNrZ3JvdW5kLWNvbG9yOiNmZmZ9YW1wLWlmcmFtZSBpZnJhbWV7Ym94LXNpemluZzpib3JkZXItYm94IWltcG9ydGFudH1bYW1wLWFjY2Vzc11bYW1wLWFjY2Vzcy1oaWRlXXtkaXNwbGF5Om5vbmV9W3N1YnNjcmlwdGlvbnMtZGlhbG9nXSxib2R5Om5vdCguaS1hbXBodG1sLXN1YnMtcmVhZHkpIFtzdWJzY3JpcHRpb25zLWFjdGlvbl0sYm9keTpub3QoLmktYW1waHRtbC1zdWJzLXJlYWR5KSBbc3Vic2NyaXB0aW9ucy1zZWN0aW9uXXtkaXNwbGF5Om5vbmUhaW1wb3J0YW50fWFtcC1leHBlcmltZW50LGFtcC1saXZlLWxpc3Q+W3VwZGF0ZV17ZGlzcGxheTpub25lfS5pLWFtcGh0bWwtamFuay1tZXRlcntwb3NpdGlvbjpmaXhlZDtiYWNrZ3JvdW5kLWNvbG9yOnJnYmEoMjMyLDcyLDk1LDAuNSk7Ym90dG9tOjA7cmlnaHQ6MDtjb2xvcjojZmZmO2ZvbnQtc2l6ZToxNnB4O3otaW5kZXg6MTAwMDtwYWRkaW5nOjVweH1hbXAtbGlzdFtyZXNpemFibGUtY2hpbGRyZW5dPi5pLWFtcGh0bWwtbG9hZGluZy1jb250YWluZXIuYW1wLWhpZGRlbntkaXNwbGF5Om5vbmUhaW1wb3J0YW50fWFtcC1saXN0IFtmZXRjaC1lcnJvcl0sYW1wLWxpc3RbbG9hZC1tb3JlXSBbbG9hZC1tb3JlLWJ1dHRvbl0sYW1wLWxpc3RbbG9hZC1tb3JlXSBbbG9hZC1tb3JlLWVuZF0sYW1wLWxpc3RbbG9hZC1tb3JlXSBbbG9hZC1tb3JlLWZhaWxlZF0sYW1wLWxpc3RbbG9hZC1tb3JlXSBbbG9hZC1tb3JlLWxvYWRpbmdde2Rpc3BsYXk6bm9uZX1hbXAtbGlzdFtkaWZmYWJsZV0gZGl2W3JvbGU9bGlzdF17ZGlzcGxheTpibG9ja31hbXAtc3RvcnktcGFnZSxhbXAtc3Rvcnlbc3RhbmRhbG9uZV17bWluLWhlaWdodDoxcHghaW1wb3J0YW50O2Rpc3BsYXk6YmxvY2shaW1wb3J0YW50O2hlaWdodDoxMDAlIWltcG9ydGFudDttYXJnaW46MCFpbXBvcnRhbnQ7cGFkZGluZzowIWltcG9ydGFudDtvdmVyZmxvdzpoaWRkZW4haW1wb3J0YW50O3dpZHRoOjEwMCUhaW1wb3J0YW50fWFtcC1zdG9yeVtzdGFuZGFsb25lXXtiYWNrZ3JvdW5kLWNvbG9yOiMyMDIxMjUhaW1wb3J0YW50O3Bvc2l0aW9uOnJlbGF0aXZlIWltcG9ydGFudH1hbXAtc3RvcnktcGFnZXtiYWNrZ3JvdW5kLWNvbG9yOiM3NTc1NzV9YW1wLXN0b3J5IC5hbXAtYWN0aXZlPmRpdixhbXAtc3RvcnkgLmktYW1waHRtbC1sb2FkZXItYmFja2dyb3VuZHtkaXNwbGF5Om5vbmUhaW1wb3J0YW50fWFtcC1zdG9yeS1wYWdlOm5vdCg6Zmlyc3Qtb2YtdHlwZSk6bm90KFtkaXN0YW5jZV0pOm5vdChbYWN0aXZlXSl7dHJhbnNmb3JtOnRyYW5zbGF0ZVkoMTAwMHZoKSFpbXBvcnRhbnR9YW1wLWF1dG9jb21wbGV0ZXtwb3NpdGlvbjpyZWxhdGl2ZSFpbXBvcnRhbnQ7ZGlzcGxheTppbmxpbmUtYmxvY2shaW1wb3J0YW50fWFtcC1hdXRvY29tcGxldGU+aW5wdXQsYW1wLWF1dG9jb21wbGV0ZT50ZXh0YXJlYXtwYWRkaW5nOjAuNXJlbTtib3JkZXI6MXB4IHNvbGlkIHJnYmEoMCwwLDAsMC4zMyl9LmktYW1waHRtbC1hdXRvY29tcGxldGUtcmVzdWx0cyxhbXAtYXV0b2NvbXBsZXRlPmlucHV0LGFtcC1hdXRvY29tcGxldGU+dGV4dGFyZWF7Zm9udC1zaXplOjFyZW07bGluZS1oZWlnaHQ6MS41cmVtfVthbXAtZnhePWZseS1pbl17dmlzaWJpbGl0eTpoaWRkZW59YW1wLXNjcmlwdFtub2RvbV17cG9zaXRpb246Zml4ZWQhaW1wb3J0YW50O3RvcDowIWltcG9ydGFudDt3aWR0aDoxcHghaW1wb3J0YW50O2hlaWdodDoxcHghaW1wb3J0YW50O292ZXJmbG93OmhpZGRlbiFpbXBvcnRhbnQ7dmlzaWJpbGl0eTpoaWRkZW59Ci8qIyBzb3VyY2VVUkw9L2Nzcy9hbXBkb2MuY3NzKi9baGlkZGVuXXtkaXNwbGF5Om5vbmUhaW1wb3J0YW50fS5pLWFtcGh0bWwtZWxlbWVudHtkaXNwbGF5OmlubGluZS1ibG9ja30uaS1hbXBodG1sLWJsdXJyeS1wbGFjZWhvbGRlcnt0cmFuc2l0aW9uOm9wYWNpdHkgMC4zcyBjdWJpYy1iZXppZXIoMC4wLDAuMCwwLjIsMSkhaW1wb3J0YW50O3BvaW50ZXItZXZlbnRzOm5vbmV9W2xheW91dD1ub2Rpc3BsYXldOm5vdCguaS1hbXBodG1sLWVsZW1lbnQpe2Rpc3BsYXk6bm9uZSFpbXBvcnRhbnR9LmktYW1waHRtbC1sYXlvdXQtZml4ZWQsW2xheW91dD1maXhlZF1bd2lkdGhdW2hlaWdodF06bm90KC5pLWFtcGh0bWwtbGF5b3V0LWZpeGVkKXtkaXNwbGF5OmlubGluZS1ibG9jaztwb3NpdGlvbjpyZWxhdGl2ZX0uaS1hbXBodG1sLWxheW91dC1yZXNwb25zaXZlLFtsYXlvdXQ9cmVzcG9uc2l2ZV1bd2lkdGhdW2hlaWdodF06bm90KC5pLWFtcGh0bWwtbGF5b3V0LXJlc3BvbnNpdmUpLFt3aWR0aF1baGVpZ2h0XVtoZWlnaHRzXTpub3QoW2xheW91dF0pOm5vdCguaS1hbXBodG1sLWxheW91dC1yZXNwb25zaXZlKSxbd2lkdGhdW2hlaWdodF1bc2l6ZXNdOm5vdChbbGF5b3V0XSk6bm90KC5pLWFtcGh0bWwtbGF5b3V0LXJlc3BvbnNpdmUpe2Rpc3BsYXk6YmxvY2s7cG9zaXRpb246cmVsYXRpdmV9LmktYW1waHRtbC1sYXlvdXQtaW50cmluc2ljLFtsYXlvdXQ9aW50cmluc2ljXVt3aWR0aF1baGVpZ2h0XTpub3QoLmktYW1waHRtbC1sYXlvdXQtaW50cmluc2ljKXtkaXNwbGF5OmlubGluZS1ibG9jaztwb3NpdGlvbjpyZWxhdGl2ZTttYXgtd2lkdGg6MTAwJX0uaS1hbXBodG1sLWxheW91dC1pbnRyaW5zaWMgLmktYW1waHRtbC1zaXplcnttYXgtd2lkdGg6MTAwJX0uaS1hbXBodG1sLWludHJpbnNpYy1zaXplcnttYXgtd2lkdGg6MTAwJTtkaXNwbGF5OmJsb2NrIWltcG9ydGFudH0uaS1hbXBodG1sLWxheW91dC1jb250YWluZXIsLmktYW1waHRtbC1sYXlvdXQtZml4ZWQtaGVpZ2h0LFtsYXlvdXQ9Y29udGFpbmVyXSxbbGF5b3V0PWZpeGVkLWhlaWdodF1baGVpZ2h0XTpub3QoLmktYW1waHRtbC1sYXlvdXQtZml4ZWQtaGVpZ2h0KXtkaXNwbGF5OmJsb2NrO3Bvc2l0aW9uOnJlbGF0aXZlfS5pLWFtcGh0bWwtbGF5b3V0LWZpbGwsLmktYW1waHRtbC1sYXlvdXQtZmlsbC5pLWFtcGh0bWwtbm90YnVpbHQsW2xheW91dD1maWxsXTpub3QoLmktYW1waHRtbC1sYXlvdXQtZmlsbCl7ZGlzcGxheTpibG9jaztvdmVyZmxvdzpoaWRkZW4haW1wb3J0YW50O3Bvc2l0aW9uOmFic29sdXRlO3RvcDowO2xlZnQ6MDtib3R0b206MDtyaWdodDowfS5pLWFtcGh0bWwtbGF5b3V0LWZsZXgtaXRlbSxbbGF5b3V0PWZsZXgtaXRlbV06bm90KC5pLWFtcGh0bWwtbGF5b3V0LWZsZXgtaXRlbSl7ZGlzcGxheTpibG9jaztwb3NpdGlvbjpyZWxhdGl2ZTstbXMtZmxleDoxIDEgYXV0bztmbGV4OjEgMSBhdXRvfS5pLWFtcGh0bWwtbGF5b3V0LWZsdWlke3Bvc2l0aW9uOnJlbGF0aXZlfS5pLWFtcGh0bWwtbGF5b3V0LXNpemUtZGVmaW5lZHtvdmVyZmxvdzpoaWRkZW4haW1wb3J0YW50fS5pLWFtcGh0bWwtbGF5b3V0LWF3YWl0aW5nLXNpemV7cG9zaXRpb246YWJzb2x1dGUhaW1wb3J0YW50O3RvcDphdXRvIWltcG9ydGFudDtib3R0b206YXV0byFpbXBvcnRhbnR9aS1hbXBodG1sLXNpemVye2Rpc3BsYXk6YmxvY2shaW1wb3J0YW50fS5pLWFtcGh0bWwtYmx1cnJ5LXBsYWNlaG9sZGVyLC5pLWFtcGh0bWwtZmlsbC1jb250ZW50e2Rpc3BsYXk6YmxvY2s7aGVpZ2h0OjA7bWF4LWhlaWdodDoxMDAlO21heC13aWR0aDoxMDAlO21pbi1oZWlnaHQ6MTAwJTttaW4td2lkdGg6MTAwJTt3aWR0aDowO21hcmdpbjphdXRvfS5pLWFtcGh0bWwtbGF5b3V0LXNpemUtZGVmaW5lZCAuaS1hbXBodG1sLWZpbGwtY29udGVudHtwb3NpdGlvbjphYnNvbHV0ZTt0b3A6MDtsZWZ0OjA7Ym90dG9tOjA7cmlnaHQ6MH0uaS1hbXBodG1sLXJlcGxhY2VkLWNvbnRlbnQsLmktYW1waHRtbC1zY3JlZW4tcmVhZGVye3BhZGRpbmc6MCFpbXBvcnRhbnQ7Ym9yZGVyOm5vbmUhaW1wb3J0YW50fS5pLWFtcGh0bWwtc2NyZWVuLXJlYWRlcntwb3NpdGlvbjpmaXhlZCFpbXBvcnRhbnQ7dG9wOjBweCFpbXBvcnRhbnQ7bGVmdDowcHghaW1wb3J0YW50O3dpZHRoOjRweCFpbXBvcnRhbnQ7aGVpZ2h0OjRweCFpbXBvcnRhbnQ7b3BhY2l0eTowIWltcG9ydGFudDtvdmVyZmxvdzpoaWRkZW4haW1wb3J0YW50O21hcmdpbjowIWltcG9ydGFudDtkaXNwbGF5OmJsb2NrIWltcG9ydGFudDt2aXNpYmlsaXR5OnZpc2libGUhaW1wb3J0YW50fS5pLWFtcGh0bWwtc2NyZWVuLXJlYWRlcn4uaS1hbXBodG1sLXNjcmVlbi1yZWFkZXJ7bGVmdDo4cHghaW1wb3J0YW50fS5pLWFtcGh0bWwtc2NyZWVuLXJlYWRlcn4uaS1hbXBodG1sLXNjcmVlbi1yZWFkZXJ+LmktYW1waHRtbC1zY3JlZW4tcmVhZGVye2xlZnQ6MTJweCFpbXBvcnRhbnR9LmktYW1waHRtbC1zY3JlZW4tcmVhZGVyfi5pLWFtcGh0bWwtc2NyZWVuLXJlYWRlcn4uaS1hbXBodG1sLXNjcmVlbi1yZWFkZXJ+LmktYW1waHRtbC1zY3JlZW4tcmVhZGVye2xlZnQ6MTZweCFpbXBvcnRhbnR9LmktYW1waHRtbC11bnJlc29sdmVke3Bvc2l0aW9uOnJlbGF0aXZlO292ZXJmbG93OmhpZGRlbiFpbXBvcnRhbnR9LmktYW1waHRtbC1zZWxlY3QtZGlzYWJsZWR7LXdlYmtpdC11c2VyLXNlbGVjdDpub25lIWltcG9ydGFudDstbW96LXVzZXItc2VsZWN0Om5vbmUhaW1wb3J0YW50Oy1tcy11c2VyLXNlbGVjdDpub25lIWltcG9ydGFudDt1c2VyLXNlbGVjdDpub25lIWltcG9ydGFudH0uaS1hbXBodG1sLW5vdGJ1aWx0LFtsYXlvdXRdOm5vdCguaS1hbXBodG1sLWVsZW1lbnQpLFt3aWR0aF1baGVpZ2h0XVtoZWlnaHRzXTpub3QoW2xheW91dF0pOm5vdCguaS1hbXBodG1sLWVsZW1lbnQpLFt3aWR0aF1baGVpZ2h0XVtzaXplc106bm90KFtsYXlvdXRdKTpub3QoLmktYW1waHRtbC1lbGVtZW50KXtwb3NpdGlvbjpyZWxhdGl2ZTtvdmVyZmxvdzpoaWRkZW4haW1wb3J0YW50O2NvbG9yOnRyYW5zcGFyZW50IWltcG9ydGFudH0uaS1hbXBodG1sLW5vdGJ1aWx0Om5vdCguaS1hbXBodG1sLWxheW91dC1jb250YWluZXIpPiosW2xheW91dF06bm90KFtsYXlvdXQ9Y29udGFpbmVyXSk6bm90KC5pLWFtcGh0bWwtZWxlbWVudCk+Kixbd2lkdGhdW2hlaWdodF1baGVpZ2h0c106bm90KFtsYXlvdXRdKTpub3QoLmktYW1waHRtbC1lbGVtZW50KT4qLFt3aWR0aF1baGVpZ2h0XVtzaXplc106bm90KFtsYXlvdXRdKTpub3QoLmktYW1waHRtbC1lbGVtZW50KT4qe2Rpc3BsYXk6bm9uZX1hbXAtaW1nOm5vdCguaS1hbXBodG1sLWVsZW1lbnQpW2ktYW1waHRtbC1zc3JdPmltZy5pLWFtcGh0bWwtZmlsbC1jb250ZW50e2Rpc3BsYXk6YmxvY2t9LmktYW1waHRtbC1ub3RidWlsdDpub3QoLmktYW1waHRtbC1sYXlvdXQtY29udGFpbmVyKSxbbGF5b3V0XTpub3QoW2xheW91dD1jb250YWluZXJdKTpub3QoLmktYW1waHRtbC1lbGVtZW50KSxbd2lkdGhdW2hlaWdodF1baGVpZ2h0c106bm90KFtsYXlvdXRdKTpub3QoLmktYW1waHRtbC1lbGVtZW50KSxbd2lkdGhdW2hlaWdodF1bc2l6ZXNdOm5vdChbbGF5b3V0XSk6bm90KC5pLWFtcGh0bWwtZWxlbWVudCl7Y29sb3I6dHJhbnNwYXJlbnQhaW1wb3J0YW50O2xpbmUtaGVpZ2h0OjAhaW1wb3J0YW50fS5pLWFtcGh0bWwtZ2hvc3R7dmlzaWJpbGl0eTpoaWRkZW4haW1wb3J0YW50fS5pLWFtcGh0bWwtZWxlbWVudD5bcGxhY2Vob2xkZXJdLFtsYXlvdXRdOm5vdCguaS1hbXBodG1sLWVsZW1lbnQpPltwbGFjZWhvbGRlcl0sW3dpZHRoXVtoZWlnaHRdW2hlaWdodHNdOm5vdChbbGF5b3V0XSk6bm90KC5pLWFtcGh0bWwtZWxlbWVudCk+W3BsYWNlaG9sZGVyXSxbd2lkdGhdW2hlaWdodF1bc2l6ZXNdOm5vdChbbGF5b3V0XSk6bm90KC5pLWFtcGh0bWwtZWxlbWVudCk+W3BsYWNlaG9sZGVyXXtkaXNwbGF5OmJsb2NrfS5pLWFtcGh0bWwtZWxlbWVudD5bcGxhY2Vob2xkZXJdLmFtcC1oaWRkZW4sLmktYW1waHRtbC1lbGVtZW50PltwbGFjZWhvbGRlcl0uaGlkZGVue3Zpc2liaWxpdHk6aGlkZGVufS5pLWFtcGh0bWwtZWxlbWVudDpub3QoLmFtcC1ub3RzdXBwb3J0ZWQpPltmYWxsYmFja10sLmktYW1waHRtbC1sYXlvdXQtY29udGFpbmVyPltwbGFjZWhvbGRlcl0uYW1wLWhpZGRlbiwuaS1hbXBodG1sLWxheW91dC1jb250YWluZXI+W3BsYWNlaG9sZGVyXS5oaWRkZW57ZGlzcGxheTpub25lfS5pLWFtcGh0bWwtbGF5b3V0LXNpemUtZGVmaW5lZD5bZmFsbGJhY2tdLC5pLWFtcGh0bWwtbGF5b3V0LXNpemUtZGVmaW5lZD5bcGxhY2Vob2xkZXJde3Bvc2l0aW9uOmFic29sdXRlIWltcG9ydGFudDt0b3A6MCFpbXBvcnRhbnQ7bGVmdDowIWltcG9ydGFudDtyaWdodDowIWltcG9ydGFudDtib3R0b206MCFpbXBvcnRhbnQ7ei1pbmRleDoxfS5pLWFtcGh0bWwtbm90YnVpbHQ+W3BsYWNlaG9sZGVyXXtkaXNwbGF5OmJsb2NrIWltcG9ydGFudH0uaS1hbXBodG1sLWhpZGRlbi1ieS1tZWRpYS1xdWVyeXtkaXNwbGF5Om5vbmUhaW1wb3J0YW50fS5pLWFtcGh0bWwtZWxlbWVudC1lcnJvcntiYWNrZ3JvdW5kOnJlZCFpbXBvcnRhbnQ7Y29sb3I6I2ZmZiFpbXBvcnRhbnQ7cG9zaXRpb246cmVsYXRpdmUhaW1wb3J0YW50fS5pLWFtcGh0bWwtZWxlbWVudC1lcnJvcjpiZWZvcmV7Y29udGVudDphdHRyKGVycm9yLW1lc3NhZ2UpfWktYW1wLXNjcm9sbC1jb250YWluZXIsaS1hbXBodG1sLXNjcm9sbC1jb250YWluZXJ7cG9zaXRpb246YWJzb2x1dGU7dG9wOjA7bGVmdDowO3JpZ2h0OjA7Ym90dG9tOjA7ZGlzcGxheTpibG9ja31pLWFtcC1zY3JvbGwtY29udGFpbmVyLmFtcC1hY3RpdmUsaS1hbXBodG1sLXNjcm9sbC1jb250YWluZXIuYW1wLWFjdGl2ZXtvdmVyZmxvdzphdXRvOy13ZWJraXQtb3ZlcmZsb3ctc2Nyb2xsaW5nOnRvdWNofS5pLWFtcGh0bWwtbG9hZGluZy1jb250YWluZXJ7ZGlzcGxheTpibG9jayFpbXBvcnRhbnQ7cG9pbnRlci1ldmVudHM6bm9uZTt6LWluZGV4OjF9LmktYW1waHRtbC1ub3RidWlsdD4uaS1hbXBodG1sLWxvYWRpbmctY29udGFpbmVye2Rpc3BsYXk6YmxvY2shaW1wb3J0YW50fS5pLWFtcGh0bWwtbG9hZGluZy1jb250YWluZXIuYW1wLWhpZGRlbnt2aXNpYmlsaXR5OmhpZGRlbn0uaS1hbXBodG1sLWVsZW1lbnQ+W292ZXJmbG93XXtjdXJzb3I6cG9pbnRlcjtwb3NpdGlvbjpyZWxhdGl2ZTt6LWluZGV4OjI7dmlzaWJpbGl0eTpoaWRkZW47ZGlzcGxheTppbml0aWFsO2xpbmUtaGVpZ2h0Om5vcm1hbH0uaS1hbXBodG1sLWVsZW1lbnQ+W292ZXJmbG93XS5hbXAtdmlzaWJsZXt2aXNpYmlsaXR5OnZpc2libGV9dGVtcGxhdGV7ZGlzcGxheTpub25lIWltcG9ydGFudH0uYW1wLWJvcmRlci1ib3gsLmFtcC1ib3JkZXItYm94ICosLmFtcC1ib3JkZXItYm94IDphZnRlciwuYW1wLWJvcmRlci1ib3ggOmJlZm9yZXtib3gtc2l6aW5nOmJvcmRlci1ib3h9YW1wLXBpeGVse2Rpc3BsYXk6bm9uZSFpbXBvcnRhbnR9YW1wLWFuYWx5dGljcyxhbXAtYXV0by1hZHMsYW1wLXN0b3J5LWF1dG8tYWRze3Bvc2l0aW9uOmZpeGVkIWltcG9ydGFudDt0b3A6MCFpbXBvcnRhbnQ7d2lkdGg6MXB4IWltcG9ydGFudDtoZWlnaHQ6MXB4IWltcG9ydGFudDtvdmVyZmxvdzpoaWRkZW4haW1wb3J0YW50O3Zpc2liaWxpdHk6aGlkZGVufWh0bWwuaS1hbXBodG1sLWZpZT5hbXAtYW5hbHl0aWNze3Bvc2l0aW9uOmluaXRpYWwhaW1wb3J0YW50fVt2aXNpYmxlLXdoZW4taW52YWxpZF06bm90KC52aXNpYmxlKSxmb3JtIFtzdWJtaXQtZXJyb3JdLGZvcm0gW3N1Ym1pdC1zdWNjZXNzXSxmb3JtIFtzdWJtaXR0aW5nXXtkaXNwbGF5Om5vbmV9YW1wLWFjY29yZGlvbntkaXNwbGF5OmJsb2NrIWltcG9ydGFudH1hbXAtYWNjb3JkaW9uPnNlY3Rpb257ZmxvYXQ6bm9uZSFpbXBvcnRhbnR9YW1wLWFjY29yZGlvbj5zZWN0aW9uPip7ZmxvYXQ6bm9uZSFpbXBvcnRhbnQ7ZGlzcGxheTpibG9jayFpbXBvcnRhbnQ7b3ZlcmZsb3c6aGlkZGVuIWltcG9ydGFudDtwb3NpdGlvbjpyZWxhdGl2ZSFpbXBvcnRhbnR9YW1wLWFjY29yZGlvbixhbXAtYWNjb3JkaW9uPnNlY3Rpb257bWFyZ2luOjB9YW1wLWFjY29yZGlvbj5zZWN0aW9uPjpsYXN0LWNoaWxke2Rpc3BsYXk6bm9uZSFpbXBvcnRhbnR9YW1wLWFjY29yZGlvbj5zZWN0aW9uW2V4cGFuZGVkXT46bGFzdC1jaGlsZHtkaXNwbGF5OmJsb2NrIWltcG9ydGFudH0KLyojIHNvdXJjZVVSTD0vY3NzL2FtcHNoYXJlZC5jc3MqLzwvc3R5bGU+CiAgPG1ldGEgbmFtZT0idmlld3BvcnQiIGNvbnRlbnQ9IndpZHRoPWRldmljZS13aWR0aCxtaW5pbXVtLXNjYWxlPTEsaW5pdGlhbC1zY2FsZT0xIj4KICA8bGluayByZWw9InByZWxvYWQiIGhyZWY9Imh0dHBzOi8vY2RuLmFtcHByb2plY3Qub3JnL3YwLmpzIiBhcz0ic2NyaXB0Ij4KICA8c2NyaXB0IGFzeW5jIHNyYz0iaHR0cHM6Ly9jZG4uYW1wcHJvamVjdC5vcmcvdjAuanMiPjwvc2NyaXB0PgogIDxsaW5rIHJlbD0icHJlbG9hZCIgaHJlZj0iaHR0cHM6Ly9hbXAuZGV2L3N0YXRpYy9pbWcvaWNvbnMvaWNvbi01MTJ4NTEyLnBuZyIgYXM9ImltYWdlIiBkYXRhLWhlcm8+CiAgPHRpdGxlPk15IEFNUCBQYWdlPC90aXRsZT4KICA8bGluayByZWw9ImNhbm9uaWNhbCIgaHJlZj0ic2VsZi5odG1sIj4KPC9oZWFkPgo8Ym9keT4KICA8YW1wLWltZyBzcmM9Imh0dHBzOi8vYW1wLmRldi9zdGF0aWMvaW1nL2ljb25zL2ljb24tNTEyeDUxMi5wbmciIHdpZHRoPSI1MTIiIGhlaWdodD0iMjUwIiBvYmplY3QtZml0PSJjb3ZlciIgb2JqZWN0LXBvc2l0aW9uPSJyaWdodCB0b3AiIGktYW1waHRtbC1zc3IgZGF0YS1oZXJvIGNsYXNzPSJpLWFtcGh0bWwtbGF5b3V0LWZpeGVkIGktYW1waHRtbC1sYXlvdXQtc2l6ZS1kZWZpbmVkIiBzdHlsZT0id2lkdGg6NTEycHg7aGVpZ2h0OjI1MHB4OyIgaS1hbXBodG1sLWxheW91dD0iZml4ZWQiPjxpbWcgY2xhc3M9ImktYW1waHRtbC1maWxsLWNvbnRlbnQgaS1hbXBodG1sLXJlcGxhY2VkLWNvbnRlbnQiIGRlY29kaW5nPSJhc3luYyIgb2JqZWN0LWZpdD0iY292ZXIiIG9iamVjdC1wb3NpdGlvbj0icmlnaHQgdG9wIiBzcmM9Imh0dHBzOi8vYW1wLmRldi9zdGF0aWMvaW1nL2ljb25zL2ljb24tNTEyeDUxMi5wbmciPjwvYW1wLWltZz4KICAKICA8YW1wLWltZyBzcmM9Imh0dHBzOi8vYW1wLmRldi9zdGF0aWMvaW1nL2ljb25zL2ljb24tNTEyeDUxMi5wbmciIHdpZHRoPSI1MTIiIGhlaWdodD0iMjUwIiBvYmplY3QtZml0PSJjb3ZlciIgb2JqZWN0LXBvc2l0aW9uPSJyaWdodCB0b3AiIGktYW1waHRtbC1zc3IgZGF0YS1oZXJvIGNsYXNzPSJpLWFtcGh0bWwtbGF5b3V0LWZpeGVkIGktYW1waHRtbC1sYXlvdXQtc2l6ZS1kZWZpbmVkIiBzdHlsZT0id2lkdGg6NTEycHg7aGVpZ2h0OjI1MHB4OyIgaS1hbXBodG1sLWxheW91dD0iZml4ZWQiPgoJPGltZyBjbGFzcz0iaS1hbXBodG1sLWZpbGwtY29udGVudCBpLWFtcGh0bWwtcmVwbGFjZWQtY29udGVudCIgZGVjb2Rpbmc9ImFzeW5jIiBzdHlsZT0ib2JqZWN0LWZpdDogY292ZXI7IG9iamVjdC1wb3NpdGlvbjogcmlnaHQgdG9wOyIgc3JjPSJodHRwczovL2FtcC5kZXYvc3RhdGljL2ltZy9pY29ucy9pY29uLTUxMng1MTIucG5nIj4KPC9hbXAtaW1nPgo8L2JvZHk+CjwvaHRtbD4=).

### LINKED ISSUE #19: Server-side rendering can cause style[amp-custom] CSS overage
## Bug Description

Given an AMP page that has an `style[amp-custom]` which is just under the 75000 byte limit at 74,998 bytes. 

If this AMP page also has an `amp-analytics` tag, then the SSR transformer will end up adding an inline `style` to this element like so:

```html
<amp-analytics type="analytics" class="i-amphtml-layout-fixed i-amphtml-layout-size-defined" style="width:1px;height:1px;" i-amphtml-layout="fixed">
```

The result is an AMP validation error is leaked to the frontend:

> The author stylesheet specified in tag ‘style amp-custom’ and the combined inline styles is too large – document contains 75023 bytes whereas the limit is 75000 bytes.

Why is this? It's because `width:1px;height:1px;` (21 bytes) is _added_ to the size of `style[amp-custom]` (74,998 bytes): 74,998 + 21 = 75,023. The key line in the error message is “and the combined inline styles”.

Originally reported in support topic: https://wordpress.org/support/topic/gcs-error/

Example document that has this problem: https://gist.github.com/westonruter/4dd1f81cc393568e9eef3e008f6615e6#file-que-son-gadget-amp-html

The current workaround is to install the [AMP Disable SSR](https://gist.github.com/westonruter/8d52c0b807e6dfbbdf2219622d0f4a7e) plugin.

## Expected Behaviour

SSR should not cause the total `amp-custom` CSS to go over 75000 bytes. It should keep track of how much CSS it is adding to the page, and if that causes the total to go over what is already in `style[amp-custom]`, it should abort SSR.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
This PR:
- Adds new `Dom\Element` abstraction that extends `DOMElement` and is returned by `Dom\Document` by default.
- Adds centralized code contained within `Dom\Document` & `Dom\Element` to keep track of the byte count across the `<style amp-custom>` tag and all inline styles.
- Adds `Dom\Document::addAmpCustomStyle()` & `Dom\Element::addInlineStyle()` that adds CSS while enforcing the maximum CSS byte count, and throwing a `MaxCssByteCountExceeded` exception when going over the limit.
- Ensures all inline style manipulations within `Optimizer\Transformer\ServerSideRendering` use the above abstractions to always enforce the CSS byte count maximum.
- Adds logic to `Optimizer\Transformer\PreloadHeroImage` to turn `object-fit` & `object-position` attributes into inline styles.

Fixes #6 
Fixes #19


src/Dom/Element.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    const PROPERTY_GETTER_ERROR_MESSAGE = 'Undefined property: AmpProject\\Dom\\Element::';
+
+    /**
+     * Add CSS styles to the element as an inline style attribute.
+     *
+     * @param string $style CSS style(s) to add to the inline style attribute.
+     * @throws MaxCssByteCountExceeded If the allowed max byte count is exceeded.
+     */
+    public function addInlineStyle($style)
```
westonruter, line 34: What if a sanitizer does `setAttribute('style', '...')`? For example: 

https://github.com/ampproject/amp-wp/blob/06c9b9531874174facfa805cad683c73617360a8/includes/sanitizers/class-amp-base-sanitizer.php#L723

Should this not be implemented to route to `addInlineStyle()` (with a recursion check of course)? 

Otherwise, it would be important that `addInlineStyle` never be called until the last `style` attribute is set directly, otherwise the `inlineStyleByteCount` will not be correct.

Another complication: if someone does `$attrNode = $element->getAttributeNode('style')` and then sets the value via `$attrNode->nodeValue`. That would be another way for the byte `inlineStyleByteCount` to become out of sync.

src/Dom/Element.php
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @param string $name Name of the property to get.
+     * @return mixed Value of the property, or null if unknown property was requested.
+     */
+    public function __get($name)
+    {
+        switch ($name) {
+            case 'inlineStyleByteCount':
+                if (!isset($this->inlineStyleByteCount)) {
+                    $this->inlineStyleByteCount = strlen((string)$this->getAttribute(Attribute::STYLE));
```
westonruter, line 67: Caching the length seems perhaps overkill and will introduce yet another area where the count can get out of sync.

src/Dom/ElementDump.php
```diff
@@ -0,0 +1,78 @@
...
+        );
+
+        if ($this->truncate > 0 && mb_strlen($attributes) > $this->truncate) {
+            $attributes = mb_substr($attributes, 0, $this->truncate - 1) . '…';
+        }
+
+        $textContent = $this->element->textContent;
+
+        if ($this->truncate > 0 && mb_strlen($textContent) > $this->truncate) {
+            $textContent = mb_substr($textContent, 0, $this->truncate - 1) . '…';
```
westonruter, line 68: Warning: The AMP plugin does not require the `mbstring` extension. It is a `suggestion`'ing. So `mb_strlen()` and `mb_substr()` may not exist.

src/Optimizer/Transformer/PreloadHeroImage.php
```diff
@@ -521,14 +529,22 @@ private function generateImg(HeroImage $heroImage, Document $document)
             }
         }
 
+        foreach (self::ATTRIBUTES_TO_INLINE as $attribute) {
+            if ($element->hasAttribute($attribute)) {
+                $value = $element->getAttribute($attribute);
+                $style = empty($value) ? '' : "{$attribute}:{$element->getAttribute($attribute)}";
+                $imgElement->addInlineStyle($style);
```
westonruter, line 536: Since this can throw an exception, shouldn't the each transformer's `transform` call be wrapped in a `try`/`catch` block in `TransformationEngine::optimizeDom()`:

https://github.com/ampproject/amp-toolbox-php/blob/c7ccb98b6bbd96ed5c5546302f272c70b03116e6/src/Optimizer/TransformationEngine.php#L62-L67

The caught `$exception` can be added to `$errors`. 

I recall we discussed this earlier today. The `MaxCssByteCountExceeded` exception would be classed as a catchable exception which doesn't necessitate abandoning optimization entirely. Or rather, perhaps this one specific call to `$imgElement->addInlineStyle($style);` should be wrapped in a `try {…} catch ( MaxCssByteCountExceeded $exception ) {…}` to then re-throw a catchable `TransformerAbortedException`.

src/Dom/Element.php
westonruter: Additionally, what if someone creates a `<style>` element and appends it to the document? The style sanitizer will process those in the AMP plugin, but once passed over to the Optimizer then custom transformers could do that. Though I guess they could also add arbitrary other invalid elements, as well as add invalid `style` attributes. So it would be the responsibility of transformers to not do bad things?

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -526,12 +484,8 @@ private function applyLayoutAttributes(DOMElement $element, $layout, CssLength $
...
         }
 
-        // We prepend just in case an existing value (which shouldn't be there for valid docs) doesn't end with ';'.
-        if ($element->hasAttribute(Tag::STYLE)) {
-            $styles .= $element->getAttribute(Tag::STYLE);
-        }
-        if (! empty($styles)) {
-            $element->setAttribute(Tag::STYLE, $styles);
+        if (!empty($styles)) {
+            $element->addInlineStyle($styles);
```
westonruter, line 488: This could also throw an exception, and when that happens, it should get caught in `\AmpProject\Optimizer\Transformer\ServerSideRendering::applyLayout()` and then added to to `$errors` to report that SSR failed to complete, right?

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -643,7 +597,7 @@ private function createResponsiveSizer(
         $style = empty($style) ? 'display:block' : "display:block;{$style}";
 
         $sizer = $document->createElement(Amp::SIZER_ELEMENT);
-        $sizer->setAttribute(Tag::STYLE, sprintf($style, $paddingString));
+        $sizer->addInlineStyle(sprintf($style, $paddingString));
```
westonruter, line 600: 〃 ☝️

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -800,26 +754,17 @@ private function adaptBlockingAttributes(Document $document, DOMElement $ampElem
                             break;
                         }
 
-                        $customCss            = array_merge(
-                            $customCss,
-                            $this->extractSizesAttributeCss($document, $ampElement, $attribute)
-                        );
+                        $customCss += $this->extractSizesAttributeCss($document, $ampElement, $attribute);
```
westonruter, line 757: Is merging arrays in this way safe? In my experience it fails to add items that have the same array keys:

```php
$x = [ 1, 2, 3 ];
$x += [ 4, 5, 6, 7 ];
print_r( $x ); 
```

```
Array
(
    [0] => 1
    [1] => 2
    [2] => 3
    [3] => 7
)
```

src/Optimizer/Transformer/ServerSideRendering.php
westonruter: In the same way as below:

```php
            if ($document->getRemainingCustomCssSpace() < $cssRule->getByteCount()) {
                $errors->add(Error\CannotRemoveBoilerplate::fromAttributesRequiringBoilerplate($ampElement));
                return false;
            }
```

westonruter (changes requested):

src/Amp.php
```diff
@@ -116,6 +116,15 @@ final class Amp
     const SIZER_ELEMENT             = 'i-amphtml-sizer';
     const INTRINSIC_SIZER_ELEMENT   = 'i-amphtml-intrinsic-sizer';
 
+    /**
+     * Maximum size of the CSS styles in bytes.
+     *
+     * @todo Max size is hard-coded for now until we ported over the generated spec into a reusable package.
```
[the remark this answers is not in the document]
schlessera: This is more about avoiding multiple sources of truth. Once we have ported the spec over, it will be easy to refer to the value within the spec, instead of within some arbitrary class constant.

src/DevMode.php
```diff
@@ -27,15 +26,11 @@ final class DevMode
...
+     * @param Document $document Document for which to check whether dev mode is active.
      * @return bool Whether the document is in dev mode.
      */
-    public static function isActiveForDocument(DOMDocument $document)
+    public static function isActiveForDocument(Document $document)
     {
-        if (! $document instanceof Document) {
-            $document = Document::fromNode($document);
-        }
-
```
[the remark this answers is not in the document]
schlessera: I think this was just copied from https://github.com/ampproject/amp-toolbox-php/blob/fix/6-object-position-fit-in-style/src/DevMode.php#L78-L81 , but the situation is a different one.

I don't think there is anything special about this method that should allow it to be called for a regular `DOMDocument`, as we are already enforcing the `Dom\Document` implementation everywhere else. Is this more likely to be called outside of the AMP plugin context? Maybe then we should think about a helper function `amp_is_dev_mode()` or similar...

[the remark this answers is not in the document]
westonruter: OK, yeah. That makes sense to be deleted then.

src/Dom/Document.php
```diff
@@ -663,52 +692,65 @@ private function normalizeDocumentStructure($content)
      */
     public function normalizeDomStructure()
     {
-
         if (! $this->documentElement) {
             $this->appendChild($this->createElement(Tag::HTML));
```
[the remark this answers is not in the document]
westonruter: I suppose this is resolved with the discussion on #30.

src/Dom/Element.php
schlessera: Regarding the use of `setAttribute('style', ...)`, I think it is irrelevant if it happens before the style sanitizer ran. So, up until that point, it is probably just wasted processing time to calculate any byte lengths.

However, what I could do is to route everything through `addInlineStyle()`, but then have a flag I can set to start enforcing at a certain point. So the style sanitizer would basically say `$document->enforce_css_max_byte_count( 75000 )` and from this point onwards, all changes would be checked for size.

The question here is mainly whether we want to enforce the byte count across our own code, or across any future third-party extensions. Given this is a library for general consumption, it seems to make sense to go that extra mile.

src/Dom/ElementDump.php
schlessera: Good point. This was already in the count for exceptions (i just moved it), but it's probably best to add hardening nevertheless.

src/Optimizer/Transformer/ServerSideRendering.php
schlessera: I tried to avoid `array_merge()` for performance reasons, but you're right, this doesn't seem reliable.

src/Dom/Element.php
westonruter: > So the style sanitizer would basically say `$document->enforce_css_max_byte_count( 75000 )` and from this point onwards, all changes would be checked for size.

Yes, I like that idea.

schlessera: I've enforced `Dom\Element::setAttribute()` now to always keep the byte count in check once the document has an enforced limit.

I didn't add any logic to the style attribute itself (the `DOMAttr`). I'm a bit wary of extended that deep in the hierarchy for performance reasons. Do you consider `setAttribute()` to be safe enough, or should we go all in? I'd suggest staying away from extending `DOMAttr` for now and see how the extension to `DOMElement` plays out first. The entire PHP DOM subsystem always seems to be up for a few surprises...

src/Optimizer/Transformer/ServerSideRendering.php
schlessera: I safeguarded both this one and the next by wrapping at a higher level and creating a clone first so we can leave the element in an uncorrupted state:
```php
        try {
            $newElement = $element->cloneNode(true);
            $this->applyLayoutAttributes($element, $layout, $width, $height);
            $this->maybeAddSizerInto($document, $element, $layout, $width, $height);
            $element->parentNode->replaceChild($newElement, $element);
        } catch (MaxCssByteCountExceeded $exception) {
            $errors->add(
                Error\CannotPerformServerSideRendering::fromMaxCssByteCountExceededException($exception, $element)
            );
            return false;
        }
```

That seems to be the cleanest solution. Anything else might catch the exceptions, but will leave half-transformed left-overs in the DOM.

src/Optimizer/Transformer/PreloadHeroImage.php
schlessera: I add try/catch clauses around the `addInlineStyle()` calls. I don't there's any point right now in catching exceptions at the transformer level.

src/Dom/Element.php
schlessera: Yes, but I estimate that this will save quite a bit of processing in some scenarios. It replaces, per inline style, a DOM traversal + value cast + arbitrarily-sized byte count with a simple property read on each recount.

src/Dom/Element.php
westonruter: Yes, I think this is fine for now. It would be unusual for someone to be manipulating a `DOMAttr` directly.

src/Dom/ElementDump.php
```diff
@@ -75,4 +69,32 @@ static function ($text, DOMAttr $attribute) {
...
         );
     }
+
+    /**
+     * Truncate the provided text if needed.
+     *
+     * @param string $text Text to truncate.
+     * @return string Potentially truncated text.
+     */
+    private function maybeTruncate($text)
```
westonruter, line 79: This name will make @amedina happy.

westonruter (approved):
Great work.

src/Dom/ElementDump.php
schlessera: Yes, I thought so too... :D

PR #30, opened 2020-11-18
Use magic $html property in normalizeHtmlAttributes()

### LINKED ISSUE #27: Undefined property: `Document::documentElement`
In the Web Stories plugin we parse the story markup using `Document::fromHtml( $markup, get_bloginfo( 'charset' ) )` for further processing.

A user has now reported an issue with that, where somehow the parsing fails, causing the `doccumentElement` property to be `null`. Here's the screenshot of the error:

Screen Shot 2020-11-12 at 1 35 18 PM

My hunch is that it's related to an encoding issue, but not sure.

Maybe it's possible  to add some hardening to the `Document` class around instances of using `$this->documentElement`, i.e. first checking whether it exists

schlessera:
It seems that `Dom\Document::normalizeHtmlAttributes()` can be reached in such a way that the `$documentElement` property is not set.

Instead of relying on that property, this PR changes `normalizeHtmlAttributes()` so that it uses the magic `$html` property instead that the `Dom\Document` provides. This already has hardening code that deals with a missing `$documentElement` property.

Fixes #27 

cc @swissspidy

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -663,7 +663,6 @@ final class Document extends DOMDocument
      */
     public function normalizeDomStructure()
     {
-
         if (! $this->documentElement) {
             $this->appendChild($this->createElement(Tag::HTML));
         } elseif (Tag::HTML !== $this->documentElement->nodeName) {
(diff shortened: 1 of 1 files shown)


swissspidy (approved):

westonruter:
Since `normalizeDomStructure()` is checking for `documemtElement` and then appends a new `html` if empty, couldn't it result in multiple `html` elements being added to a document?

westonruter:
Also, sanitizers are directly accessing `documentElement`. If it is not set, then this should be an exception somewhere.

schlessera:
> Since normalizeDomStructure() is checking for documemtElement and then appends a new html if empty, couldn't it result in multiple html elements being added to a document?

The `documentElement` _is_ the `<html>` tag. From https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement:
> For any non-empty HTML document, documentElement will always be an <html> element.

I don't see how we could end up with multiple `<html>` tags. Can you elaborate what scenario you're thinking about?

schlessera:
> Also, sanitizers are directly accessing documentElement. If it is not set, then this should be an exception somewhere.

The above happens when you use `Dom\Document::loadHtml()` or `Dom/Document::loadHtmlFragment()`.

The sanitizers should always receive a previously assembled `Dom\Document` where one of the above methods has been called, so they should be safe.

However, it might make sense to use `$document->html` within the sanitizers for consistency and further hardening.

westonruter:
> I don't see how we could end up with multiple `<html>` tags. Can you elaborate what scenario you're thinking about?

Well, I was thinking of the case where the `documentElement` is undefined for some reason, namely #27. But that is due to the `dom` being misconfigured.

But a page can end up with multiple `html` tags, like:

```php
<?php
$dom = new DOMDocument();
$dom->loadHTML( '<html><div>one</div></html><html><div>two</div></html>' );
```

```
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body><div>one</div></body><html><div>two</div></html></html>
```

Granted this wouldn't be a problem here since `documentElement` would still point to the first `html` element.

westonruter (approved):

PR #36, opened 2020-11-26
Skip escaped slashes when normalizing self-closing tags

Alexays:
Hi!
We encountered a problem on our site, when for example for a live article, we'll have a json+ld script part like this:
```json
"articleBody":"<p><img class=\"aligncenter size-large wp-image-807214\" src=\"https:\/\/images.frandroid.com\/wp-content\/uploads\/2020\/11\/sonos-beam-1200x799.jpg\" alt=\"\" width=\"1200\" height=\"799\" \/><br \/><\/p>\n"
```
Let's take this simple `<br \/>`, the WordPress amp plugin will transform it to `<br \>` which cause a Google rich result validation errors, same thing for the `img` tag.

So, here is a fix that handle this sanitized slash and remove it while removing the normal self-closing slash.
Hope that's OK for you!
Cheers.

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -880,7 +880,7 @@ final class Document extends DOMDocument
         static $regexPattern = null;
 
         if (null === $regexPattern) {
-            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*\/)?>(?!</\1>)#';
+            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(\/|\\\/))?>(?!</\1>)#';
         }
 
         $this->selfClosingTagsTransformed = true;
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -880,7 +880,7 @@ final class Document extends DOMDocument
         static $regexPattern = null;
 
         if (null === $regexPattern) {
-            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(\/|\\\/))?>(?!</\1>)#';
+            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(?<!\\\\)\/)?>(?!</\1>)#';
         }
 
         $this->selfClosingTagsTransformed = true;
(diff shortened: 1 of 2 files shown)


westonruter:
It seems the more fundamental problem is that we are trying to fix up markup with `replaceSelfClosingTags` that is not actual markup, but rather string literals inside of text content.

Alexays:
Yep, But I didn't know enough about the codebase to fix at that level, 
Any ideas on how to fix it properly?

schlessera:
I don't think we can properly fix the root cause here at the regular expression level.

I suggest fixing the more immediate issue right now by skipping `<script></script>` tags entirely.

Here's an example of how this could be done: https://regex101.com/r/2XfMzy/1

src/Dom/Document.php
```diff
@@ -880,7 +880,7 @@ private function replaceSelfClosingTags($html)
         static $regexPattern = null;
 
         if (null === $regexPattern) {
-            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*\/)?>(?!</\1>)#';
+            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(\/|\\\/))?>(?!</\1>)#';
```
westonruter, line 883: Here's an alternative proposal:

What if the pattern is updated to prevent matching self-closing slashes that are preceded by a backslash?

```suggestion
            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(?<!\\\\)\/)?>(?!</\1>)#';
```

Alexays: LGTM, that still fix the issue and without altering the jsonld markup 🎉

Alexays:
@westonruter Your suggestion commit seem to break the CI 😢

westonruter:
@schlessera The test failure actually looks the same to what @pierlon encountered in https://github.com/ampproject/amp-wp/pull/5637.

(So it may be unrelated to the change here.)

westonruter:
@Alexays  if you update with the latest from `main` the test failure should now be fixed due to https://github.com/ampproject/amp-toolbox-php/pull/39 being merged.

Alexays:
Done @westonruter :)

westonruter:
Great. Now the tests pass. Could you also add a test case in `tests/Dom/DocumentTest.php` among the `dataDomDocument` dataset to include a Schema.org metadata `script` with your example content so that we can verify that the behavior is working as expected?

For that matter, it appears there is no test case for `replaceSelfClosingTags` generally. Either that can be added as part of this PR or as part of another.

Alexays:
Added both schema metadata script test and simple self-closing tag test @westonruter

westonruter (approved):

schlessera (approved):

schlessera:
Thanks for the PR, @Alexays !

PR #38, opened 2020-11-29
issue#37: change padding calculation from round to number_format

### LINKED ISSUE #37: numberFormatting in responsiveSizer
I encountered an problem in optimizeHtml while in an de_DE.UTF-8 environment. 
The i-amphtml-sizer's padding-top is set with comma instead of dot leading to non-rendered images.

please consiger using
        $paddingString = (string) number_format($padding, 4, '.');
instead of 
        $paddingString = (string) round($padding, 4);
in ServerSideRendering::createResponsiveSizer

SimonKraus:
round is affected by the server locale. on de_DE.UTF8 for example the padding was used with comma instead of dot which prevented images from being rendered.

Fixes #37

### DIFF
--- a/src/Optimizer/Transformer/ServerSideRendering.php
+++ b/src/Optimizer/Transformer/ServerSideRendering.php
@@ -605,7 +605,7 @@ final class ServerSideRendering implements Transformer
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = (string) number_format($padding, 4, '.');
+        $paddingString = (string) number_format($padding, 4, '.', '');
 
         $style = empty($style) ? 'display:block' : "display:block;{$style}";
 
--- a/src/Optimizer/Transformer/ServerSideRendering.php
+++ b/src/Optimizer/Transformer/ServerSideRendering.php
@@ -605,7 +605,7 @@ final class ServerSideRendering implements Transformer
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = sprintf('%.6g', $padding);
+        $paddingString = preg_replace('/\.?0*$/', '', sprintf('%.4F', $padding));
 
         $style = empty($style) ? 'display:block' : "display:block;{$style}";
 
--- a/src/Optimizer/Transformer/ServerSideRendering.php
+++ b/src/Optimizer/Transformer/ServerSideRendering.php
@@ -605,7 +605,7 @@ final class ServerSideRendering implements Transformer
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = (string) number_format($padding, 4, '.', '');
+        $paddingString = sprintf('%.6g', $padding);
 
         $style = empty($style) ? 'display:block' : "display:block;{$style}";
 
--- a/src/Optimizer/Transformer/ServerSideRendering.php
+++ b/src/Optimizer/Transformer/ServerSideRendering.php
@@ -605,7 +605,7 @@ final class ServerSideRendering implements Transformer
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = (string) round($padding, 4);
+        $paddingString = (string) number_format($padding, 4, '.');
 
         $style = empty($style) ? 'display:block' : "display:block;{$style}";
 
--- a/src/Optimizer/Transformer/ServerSideRendering.php
+++ b/src/Optimizer/Transformer/ServerSideRendering.php
@@ -605,7 +605,7 @@ final class ServerSideRendering implements Transformer
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = preg_replace('/\.?0*$/', '', sprintf('%.4F', $padding));
+        $paddingString = rtrim(rtrim(sprintf('%.4F', round($padding, 4)), '0'), '.');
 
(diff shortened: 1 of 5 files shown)


src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -605,7 +605,7 @@ private function createResponsiveSizer(
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = (string) round($padding, 4);
+        $paddingString = (string) number_format($padding, 4, '.');
```
schlessera, line 608: You need to provide the fourth argument as well here.
See https://3v4l.org/clDrF

```suggestion
        $paddingString = (string) number_format($padding, 4, '.', '');
```

schlessera (changes requested):
This needs a small fix still (see change request).

schlessera:
@SimonKraus You need to also sign the CLA (as linked to above), as this is your first contribution to this repository.

SimonKraus:
added the optional param. 
first commit was done using a different account. 
signed CLA with my github account.

schlessera:
Oh, the tests show another problem now. `round()` gets rid of unneeded trailing zeros, whereas `number_format()` seems to just stick to the requested precision no matter what.

How about using `sprintf('%.4g', $padding)` instead? I think that should produce the desired output without being prone to locale adaptations.

SimonKraus:
the 4g used two decimals (because of the two leading numbers) so i used 6g.
tests seem to be green so far, should work now.

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -605,7 +605,7 @@ private function createResponsiveSizer(
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = (string) round($padding, 4);
+        $paddingString = sprintf('%.6g', $padding);
```
schlessera, line 608: Hmm, I think this is still risky, as a number like '3.33333333' will produce 5 decimals instead of 4.

Okay, so here's an algorithm that should work:
```suggestion
        $paddingString = preg_replace('/\.?0*$/', '', sprintf('%.4F', $padding));
```
We're splitting the operation up into two separate steps, as PHP doesn't seem to provide a single function that solves all our requirements. We first round to a maximum of 4 decimals, and then we strip any unneeded trailing zeros and possibly the separator as well.

So for the back and forth, I would have assumed PHP provides a function to solve this issue, but I can't find one it seems.

schlessera (changes requested):

SimonKraus:
should do the trick

schlessera (approved):

SimonKraus:
thinking the easiest way would have been
`$paddingString = str_replace(',', '.', (string)round($padding, 4))`

schlessera:
No, that will just break for any other separator (like a space, for example).

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -605,7 +605,7 @@ private function createResponsiveSizer(
         $style = 'padding-top:%s%%'
     ) {
         $padding       = $height->getNumeral() / $width->getNumeral() * 100;
-        $paddingString = (string) round($padding, 4);
```
heiglandreas, line 608: Why not do a `str_replace(',', '.', (string)round(...));`?

schlessera: Because the conversion to string uses the locale for the conversion. So, for example, the `,` could be used for separating the thousands.

heiglandreas: Why are the trailing zeros an issue in the first place? And what about applying a `trim($numberWithTrailingZeros, '0')`?

schlessera: This is part of an "optimizer" tool. So leaving around useless additional bytes in the markup is not great as part of the "optimization" process.

And indeed, doing 2 `rtrim()` operations seems to be more performant than the regex, as shown here: https://twitter.com/JanSlabon/status/1333402043915128833

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -605,7 +605,7 @@ private function createResponsiveSizer(
...
+        $paddingString = preg_replace('/\.?0*$/', '', sprintf('%.4F', $padding));
```
schlessera, line 608: The PHP community chimed in here: https://twitter.com/schlessera/status/1333330595179532288

The regex can be replaced by a faster 2x `rtrim()` operation instead.
Also, the `sprintf()` does not round the number, so to keep the previous behavior intact, we need to `round()`  first.

```suggestion
        $paddingString = rtrim(rtrim(sprintf('%.4F', round($padding, 4)), '0'), '.');
```

schlessera (changes requested):
Alas, I have another change request, based on feedback from the PHP community. Let me know if you're starting to get fed up... ;)

src/Optimizer/Transformer/ServerSideRendering.php
heiglandreas: you should even be able to only use 1 `rtrim` with `'0.'` as the second parameter...

schlessera: That would actually strip significant zeros on the left side of the period as well, i.e. `100.0000` would be trimmed into `1`.

heiglandreas: Dammit. You're right!

schlessera:
@SimonKraus I force-pushed a rebase to get rid of the single commit that was not using a proper email to get around the CLA issue.

schlessera (approved):

schlessera:
Thanks for the PR, @SimonKraus!

PR #39, opened 2020-12-03
Fix broken spec test for converting intrinsic layout

schlessera:
The spec test for converting the `intrinsic` layout into a server-side-rendered SVG image tag is failing, as we convert it into a base64-encoded file, whereas the NodeJS version leaves the raw SVG in place (see https://github.com/ampproject/amp-toolbox/issues/844).

This PR skips the test, but also fixes the `assertSimilarMarkup()` tag-splitting logic, which obfuscated the root cause.


schlessera:
@pierlon This is the broken test you flagged during your PHP8 testing. It is not related to PHP8, it just appeared as this particular spec test was moved from `experimental` to `valid` and thus enforced by the code from that point on.

westonruter (approved):

PR #40, opened 2020-12-03
Improve PHPUnit integration

### LINKED ISSUE #41: Improve PHPUnit integration
To allow for proper testing of PHP 8 compatibility and reduce the current hassle of fetching a fixed PHPUnit version based on the PHP version, the following changes should be made:
- Switch from a globally installed PHPUnit version to one that is fetched via Composer and thus respects the platform requirements automatically
- Include https://github.com/Yoast/PHPUnit-Polyfills and write tests against the PHPUnit 9 API
- Remove custom test compact code like https://github.com/ampproject/amp-toolbox-php/blob/0.1.0/tests/src/AssertContainsCompatibility.php and switch to the polyfills provided by above library, to reduce some of the maintenance effort
- Add PHP 8 to the testing matrix
- (?>Fix PHP 8 booboos)+

schlessera:
Fixes #41


schlessera:
The stricter type system in PHP 8 already uncovered a bug in the `ImageDimensions::isTiny()` logic. :)

composer.json
```diff
@@ -16,18 +16,17 @@
...
+    "yoast/phpunit-polyfills": "^0.2.0"
   },
   "suggest": {
     "ext-json": "Provides native implementation of json_encode()/json_decode().",
     "ext-mbstring": "Used by Dom\\Document to convert encoding to UTF-8 if needed."
   },
   "config": {
-    "platform": {
-      "php": "5.6"
-    },
```
pierlon, line 30: I suppose removing this allows for PHPUnit versions > 6.0 to be installed. Is there any other way we could enforce PHP 5.6 code compatibility.? I'm not sure if that's possible wit PHPStan.

schlessera: The platform configuration is irrelevant here, as it is a library, not an end project. We don't commit the lock file, and the platform setting will be ignored when the library is not the root project.

Libraries are only meant to provide constraints from within which they are supposed to work. And we already do so above in the `require` section.

This `platform` setting would be important for the plugin, though, where we actually rely on the result of the Composer resolution, instead of only defining the constraints. As the plugin in its _built_ form is meant to be compatible from PHP 5.6 all the way up to PHP 8, we have to only use code that runs on the lowest possible common constraints. That's where the `platform` setting is then used, so that you don't build against the PHP version you're running Composer with, but you built against the lowest version of PHP that is still meant to be able to run the built artifact.

schlessera: Note that, as we have removed the platform setting now, the library will be built with different versions of its dependencies on different versions of the PHP testing matrix. And that is to be expected.

We should probably also add a `composer update --prefer-lowest` version to the test matrix, as Composer by default will prefer the highest available version.

pierlon: > The platform configuration is irrelevant here, as it is a library, not an end project. We don't commit the lock file, and the platform setting will be ignored when the library is not the root project.

Ah I see, thanks for the explanation!

tests/AmpTest.php
```diff
@@ -1,15 +1,11 @@
 <?php
 
-namespace AmpProject\Common;
+namespace AmpProject;
```
pierlon, line 3: Shouldn't this file be in `tests/AmpProject`?

pierlon (changes requested):

tests/AmpTest.php
schlessera: The tests themselves are not autoloaded. The dev-autoloader that we're using points to the `src` subfolder, and is used for complementary files that are needed as dependencies of the tests, but the test files are directly loaded by PHPUnit.

Here, the namespace is only used to define in which "context" we want to run the tests, so we don't need huge lists of imports for the tested itself.

pierlon: Got it, thanks :+1:.

src/Optimizer/ImageDimensions.php
```diff
@@ -196,9 +206,11 @@ public function getHeight()
         if ($this->height === null) {
             $this->height = -1;
             $height       = $this->image->getAttribute(Attribute::HEIGHT);
-            if (! empty($height)) {
+            if (trim($height) !== '') {
                 if (is_numeric($height)) {
-                    $this->height = $height + 0; // Cast string into either int or float as needed.
+                    $intHeight    = (int)$height;
+                    $floatHeight  = (float)$height;
+                    $this->height = $intHeight == $floatHeight ? $intHeight : $floatHeight;
```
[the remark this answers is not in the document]
schlessera: Yes, this is the sanest way I could come up with so far to convert a string into a number.

It sounds silly, but PHP actually does not provide a way to "convert string into number", you can only convert into a float _or_ an int.

So here I use a weak comparison to find out whether using a float would add anything meaningful to the number as compared to using an int.

PR #42, opened 2020-12-04
Raise PHPStan level to 5

### LINKED ISSUE #16: Raise PHPStan level to 5
## Feature description

The Common package (`ampproject/optimizer`) is currently sitting at and enforcing level 4 of PHPStan static analysis.

There is 1 error that needs to be fixed so that we can raise the level to 5 and enforce this new quality level in Travis CI.

---------------

## Acceptance criteria

* PHPStan does not show any detected errors when run on level 5.

## Implementation brief

* Change the level to `5` in the file `phpstan.neon.dist`.
* Run `composer analyze`.
* Fix detected errors until no more are found.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
This PR raises the PHPStan level to 5 and fixes the following issues:
- uses string `'0'` instead of int `0` as padding string
- uses `0` instead of `null` as default exception code
- uses `DOMDocument::createAttribute()` to create attributes without a value

Fixes #16

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -1768,7 +1768,7 @@ final class Document extends DOMDocument
                 $ampCustomStyle = $this->xpath->query(self::XPATH_AMP_CUSTOM_STYLE_QUERY, $this->head)->item(0);
                 if (!$ampCustomStyle instanceof Element) {
                     $ampCustomStyle = $this->createElement(Tag::STYLE);
-                    $ampCustomStyle->setAttribute(Attribute::AMP_CUSTOM, null);
+                    $ampCustomStyle->appendChild($this->createAttribute(Attribute::AMP_CUSTOM));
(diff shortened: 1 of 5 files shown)


src/Dom/Document.php
```diff
@@ -1768,7 +1768,7 @@ public function __get($name)
                 $ampCustomStyle = $this->xpath->query(self::XPATH_AMP_CUSTOM_STYLE_QUERY, $this->head)->item(0);
                 if (!$ampCustomStyle instanceof Element) {
                     $ampCustomStyle = $this->createElement(Tag::STYLE);
-                    $ampCustomStyle->setAttribute(Attribute::AMP_CUSTOM, null);
```
westonruter, line 1771: TIL: Using `document.createAttribute()` to bypass needing to supply a value.

westonruter (approved):

PR #49, opened 2020-12-11
Support dimensions with units

### LINKED ISSUE #4: Support unit-based width/height attributes in ImageDimensions
This is moved over from outstanding issues on the PR in the `ampproject/amp-wp` repository (https://github.com/ampproject/amp-wp/pull/5350#discussion_r515668520).

Via @westonruter:

> Even though the docs say they should be integer pixel values, this is not enforced and non-integers can be supplied.
> 
> For floats, see: ampproject/amphtml#27528
> 
> And this is valid AMP:
> 
> ```
> <amp-img src="https://amp.dev/static/img/icons/icon-512x512.png" width="50vw" height="50vh"></amp-img>
> ```
> The use of decimals and/or CSS units been allowed accidentally, but since it is valid it cannot be removed.

schlessera:
This PR adds knwoledge about dimensions units and adds comparison logic for relative units to the `ImageDimension::isTiny()` check.

Fixes #4


schlessera:
Given the reason of why this is even supported in the first place, I'm unsure how far we want to take this. The current logic will probably work correctly for the most common cases, but it is certainly not fool-proof. But I think it is hardly worth the effort to go into more detail for now. We can always fix actual issues if and when they arise.

src/Optimizer/ImageDimensions.php
```diff
@@ -119,23 +129,46 @@ public function isTiny($threshold = self::TINY_THRESHOLD)
...
+                && in_array($this->getWidthUnit(), LengthUnit::RELATIVE_UNITS, true)
+                && (is_numeric($this->getHeight()) && $this->getHeight() < $threshold)
+            ) || (
+                $this->hasHeightUnit()
+                && in_array($this->getHeightUnit(), LengthUnit::RELATIVE_UNITS, true)
+                && (is_numeric($this->getWidth()) && $this->getWidth() < $threshold)
+            )
+        ) {
+            return false;
+        }
```
westonruter, line 158: It this missing a check to see if _both_ the `width` and `height` are relative?
```suggestion
        // If relative units are in use, we cannot assume much about the final dimensions.
        $isRelativeWidth = in_array($this->getWidthUnit(), LengthUnit::RELATIVE_UNITS, true);
        $isRelativeHeight = in_array($this->getHeightUnit(), LengthUnit::RELATIVE_UNITS, true);
        if (
            (
                $isRelativeWidth && $isRelativeHeight
            ) || (
                $isRelativeWidth
                && (is_numeric($this->getHeight()) && $this->getHeight() < $threshold)
            ) || (
                $isRelativeHeight
                && (is_numeric($this->getWidth()) && $this->getWidth() < $threshold)
            )
        ) {
            return false;
        }
```

This could be optimized a bit better to

src/Optimizer/ImageDimensions.php
```diff
@@ -119,23 +129,46 @@ public function isTiny($threshold = self::TINY_THRESHOLD)
...
             (is_numeric($this->getWidth()) && $this->getWidth() <= 0)
             || (is_numeric($this->getHeight()) && $this->getHeight() <= 0)
         ) {
             return true;
         }
 
+        // If relative units are in use, we cannot assume much about the final dimensions.
+        if (
+            (
+                $this->hasWidthUnit()
```
westonruter, line 148: It doesn't seem that `$this->hasWidthUnit()` is necessary since the regex will be run in `getWidthUnit`.

src/Optimizer/ImageDimensions.php
```diff
@@ -119,23 +129,46 @@ public function isTiny($threshold = self::TINY_THRESHOLD)
...
+                && in_array($this->getWidthUnit(), LengthUnit::RELATIVE_UNITS, true)
+                && (is_numeric($this->getHeight()) && $this->getHeight() < $threshold)
+            ) || (
+                $this->hasHeightUnit()
+                && in_array($this->getHeightUnit(), LengthUnit::RELATIVE_UNITS, true)
```
westonruter, line 153: It doesn't seem that `$this->hasHeightUnit()` is necessary since the regex will be run in `getHeightUnit`.

src/Optimizer/ImageDimensions.php
```diff
@@ -162,6 +195,30 @@ public function hasHeight()
...
+     *
+     * @return bool Whether the height has a unit.
+     */
+    public function hasHeightUnit()
+    {
+        $height = $this->getHeight();
+
+        return is_string($height) && preg_match(self::UNIT_REGEX_PATTERN, $height);
+    }
+
```
westonruter, line 221: Per above, I think these methods can be eliminated in favor of just checking of the get methods return a non-empty string.

westonruter (changes requested):

src/LengthUnit.php
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Set of known absolute units.
+     *
+     * @var string[]
+     */
+    const ABSOLUTE_UNITS = [
+        self::CM,
+        self::MM,
+        self::Q,
```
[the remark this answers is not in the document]
westonruter: (No action required. Just an observation.)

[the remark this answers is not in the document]
schlessera: Yes, noticed that as well. Not sure what the deal is with Q, I sure hope we don't have ourselves another troublemaker on board... !

src/Optimizer/ImageDimensions.php
schlessera: Text seems cut off here. I'm not sure what this is supposed to be improve, as the first condition would already be triggered in that case as well (without doing both of the calculations).

src/Optimizer/ImageDimensions.php
schlessera: Removed via https://github.com/ampproject/amp-toolbox-php/pull/49/commits/b66999c3816b115d408c69fb9451b655093cbda8

src/Optimizer/ImageDimensions.php
schlessera: Remove via https://github.com/ampproject/amp-toolbox-php/pull/49/commits/b66999c3816b115d408c69fb9451b655093cbda8

src/Optimizer/ImageDimensions.php
schlessera: Removed via https://github.com/ampproject/amp-toolbox-php/pull/49/commits/b66999c3816b115d408c69fb9451b655093cbda8

src/Optimizer/ImageDimensions.php
schlessera: Ah, I assume you mean that this would allow to fail without doing the comparison against the threshold.

If so, I would assume that the most costly operation here is actually the `in_array()`. If we'd want to further optimize this piece of logic, we should first look into turning this into a hashed lookup via `array_key_exists()` instead.

src/LengthUnit.php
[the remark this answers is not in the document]
westonruter: Haha! As long as the troublemaker isn't _anonymous_.

src/Optimizer/ImageDimensions.php
westonruter: Sorry for the truncation.

If not for the optimization, what seems to me missing is adding a first condition to check if _both_ the width and height are relative. Or is that accounted for?

src/Optimizer/ImageDimensions.php
```diff
@@ -119,23 +129,44 @@ public function isTiny($threshold = self::TINY_THRESHOLD)
...
+        // If one or both of the dimensions are zero, the entire image will be invisible.
         if (
             (is_numeric($this->getWidth()) && $this->getWidth() <= 0)
             || (is_numeric($this->getHeight()) && $this->getHeight() <= 0)
         ) {
             return true;
         }
 
+        // If relative units are in use, we cannot assume much about the final dimensions.
+        if (
```
westonruter, line 146:
```suggestion
        if (
            (
                in_array($this->getWidthUnit(), LengthUnit::RELATIVE_UNITS, true)
                && in_array($this->getHeightUnit(), LengthUnit::RELATIVE_UNITS, true)
            ) || (
```

src/Optimizer/ImageDimensions.php
westonruter: This is what I'm referring to: https://github.com/ampproject/amp-toolbox-php/pull/49/files#r541983412,

schlessera: Ah, yes, I got what you mean now. The two conditions only work exclusively on a single relative unit as the other needs to be numeric. I'll adapt.

schlessera:
I added a bit of logic to actually support basic unit conversions. This will be useful for some units, especially the most obvious one: `px`.

src/Optimizer/ImageDimensions.php
schlessera: Resolved via https://github.com/ampproject/amp-toolbox-php/pull/49/commits/b76f3fce804115b085e7b50702bafb901dd776a9

src/Optimizer/ImageDimensions.php
schlessera: Resolved via https://github.com/ampproject/amp-toolbox-php/pull/49/commits/b76f3fce804115b085e7b50702bafb901dd776a9

src/Optimizer/ImageDimensions.php
```diff
@@ -129,32 +143,56 @@ public function isTiny($threshold = self::TINY_THRESHOLD)
...
         }
 
-        // If relative units are in use, we cannot assume much about the final dimensions.
+        $widthUnit  = $this->getWidthUnit();
+        $heightUnit = $this->getHeightUnit();
+
+        // Try to convert absolute units into their equivalent pixel value.
+        if (!empty($widthUnit)) {
+            $numericWidth = $this->getNumericWidth();
+            if ($numericWidth) {
```
westonruter, line 168:
```suggestion
            if (false !== $numericWidth) {
```

src/Optimizer/ImageDimensions.php
```diff
@@ -129,32 +143,56 @@ public function isTiny($threshold = self::TINY_THRESHOLD)
...
+                $width     = $numericWidth;
+                $widthUnit = '';
+            }
+        }
+        if (!empty($heightUnit)) {
+            $numericHeight = $this->getNumericHeight();
+            if ($numericHeight) {
```
westonruter, line 175:
```suggestion
            if (false !== $numericHeight) {
```

src/LengthUnit.php
```diff
@@ -172,4 +172,48 @@ interface LengthUnit
...
+
+    /**
+     * Convert a unit-based length into a number of pixels.
+     *
+     * @param int|float $value Value to convert.
+     * @param string    $unit  Unit of the value.
+     * @return int|float|false Converted value, or false if it could not be converted.
+     */
+    public static function convertIntoPixels($value, $unit)
+    {
```
westonruter, line 198: Special case since `0` is `0` regardless of unit.

```suggestion
    {
        if ( 0 === $value ) {
            return 0;
        }
```

src/Optimizer/ImageDimensions.php
westonruter: Since `0` is falsey.

westonruter (changes requested):

src/LengthUnit.php
```diff
@@ -172,4 +172,48 @@ interface LengthUnit
...
+                return $value * self::PPI / self::CM_PER_IN / 40;
+            case self::IN:
+                return $value * self::PPI;
+            case self::PC:
+                return $value * self::PPI / 6;
+            case self::PT:
+                return $value * self::PPI / 72;
+            case self::PX:
+                // No conversion needed for pixel values.
+                return $value;
```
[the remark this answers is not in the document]
schlessera: Given that this switch case will be turned into a jump table (not a sequential list of checks) starting with PHP 7.2 onwards, I thought it might be preferable to optimize for consistency (i.e. doing all the occurrences of these in alphabetical order).

Let me know if you think we should optimize for PHP lower than 7.2 instead.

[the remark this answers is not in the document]
westonruter: Cool. Optimize for the future.

westonruter (approved):
LGTM other than the unit test failure which I'm guessing is due to the spec files needing to be updated.

schlessera:
The test failure is due to https://github.com/ampproject/amp-toolbox/issues/1123 and should be fixed by https://github.com/ampproject/amp-toolbox/pull/1124.

Issue #53, opened 2021-01-26
replaceMustacheTemplateTokens: Unterminated entity reference when the mustache template contains special character


buihoanghai:
I try to apply amp-toolbox for our website then facing this issue
![image](https://user-images.githubusercontent.com/6602376/105868406-bb2a4500-6028-11eb-9623-2b7a0c3eb29b.png)

The mustache templace contains "&" character in the href
![image](https://user-images.githubusercontent.com/6602376/105869337-b4e89880-6029-11eb-9c97-d87ab1fcd192.png)

westonruter:
If you replace `&` with `&amp;` in your source code, does that prevent the error from happening?

Technically `&` alone is not valid markup, although all browsers support it and automatically parse it as `&amp;`.

westonruter:
I can reproduce the issue with this markup:

```html
<template type="amp-mustache">
    <a href="https://example.com/?foo={{foo}}&bar={{bar}}&amp;baz={{baz}}">Link</a>
</template>
```

schlessera:
It looks like this is a known bug in PHP which will not be fixed because of BC: https://bugs.php.net/bug.php?id=31191

PR #54, opened 2021-01-26
Secure HTML entities within template text

buihoanghai:
Fixes https://github.com/ampproject/amp-toolbox-php/issues/53

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -1311,12 +1311,11 @@ final class Document extends DOMDocument
                     $count
                 );
 
-                // Make sure entities are encoded setting the attribute value. Note this is at variance with browser DOM,
-                // where normally an entity supplied as a nodeValue "&amp;" would get escaped when serialized
-                // (e.g. "&amp;amp;"). In contrast, PHP DOM will try to parse the entity when it is supplied.
-                $value = htmlspecialchars( $value, ENT_QUOTES | ENT_HTML5, 'UTF-8', false );
-
-                $attribute->nodeValue = $value;
+                // Note we cannot do `$attribute->nodeValue = $value` because the PHP DOM will try to parse any
+                // entities. In the case of a URL value like '/foo/?bar=1&baz=2' the result is a warning for an
+                // unterminated entity reference "baz". When the attribute value is updated via setAttribute() this
+                // same problem does not occur, so that is why the following is used.
+                $attribute->parentNode->setAttribute( $attribute->nodeName, $value );
 
                 if ($count) {
                     $this->mustacheTagsReplaced = true;
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -525,7 +525,7 @@ final class Document extends DOMDocument
      *
      * @link  https://php.net/manual/domdocument.savehtml.php
      *
-     * @param DOMNode $node Optional. Parameter to output a subset of the document.
+     * @param DOMNode|null $node Optional. Parameter to output a subset of the document.
      * @return string The HTML, or false if an error occurred.
      */
     public function saveHTML(DOMNode $node = null)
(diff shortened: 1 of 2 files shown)


westonruter:
It doesn't make sense to me why this would work. The `nodeValue` should contain the parsed HTML form. Also, `nodeValue` is supposed to be the same as `textContent`: https://www.php.net/manual/en/class.domnode.php#domnode.props.nodevalue

Nevertheless, it behaves differently: https://3v4l.org/qr5AD

```php
<?php
$dom = new DOMDocument();
$dom->loadHTML( '<!DOCTYPE html><html></html>' );

$tests = [
	'nodeValueUnescaped'   => [ 'nodeValue', 'http://example.com/?foo&bar' ],
	'textContentUnescaped' => [ 'textContent', 'http://example.com/?foo&bar' ],
	'nodeValueEscaped'     => [ 'nodeValue', 'http://example.com/?foo&amp;bar' ],
	'textContentEscaped'   => [ 'textContent', 'http://example.com/?foo&amp;bar' ],
];

foreach ( $tests as $i => list( $prop, $value ) ) {
	$dom->documentElement->appendChild( $dom->createTextNode( "\n" ) );
	$el = $dom->createElement( 'a' );
	$el->setAttribute( 'id', $i );

	$attr = $dom->createAttribute( 'href' );
	$attr->$prop = $value;
	$el->setAttributeNode( $attr );

	$dom->documentElement->appendChild( $el );
}
$dom->documentElement->appendChild( $dom->createTextNode( "\n" ) );

echo $dom->saveHTML();
```

There are also different results in PHP 5.6.1-5.6.13 versus all other versions:

> image

Compare this to the equivalent logic in JS (with same output in Chrome and Firefox): https://codepen.io/westonruter/pen/RwGXyNE

> image

This is what I expected to see from PHP, where the `nodeValue` and `textContent` behave identically, and also where any value supplied for the property is not parsed in any way.

It seems libxml in PHP is behaving very differently than the DOM spec requires.

schlessera:
@buihoanghai The two test failures are unrelated and should be fixed in latest `main` branch. Can you please rebase this PR to latest `main` branch?

schlessera:
@buihoanghai As you can see from the test I added, the simple use of `htmlspecialchars()` here is not a complete fix, as we are already losing some of the information at an earlier transformation, and thus an already encoded entity is erroneously reduced as well. I'll have to investigate further.

schlessera:
@buihoanghai I've replaced the `htmlspecialchars` with a different approach that seems to keep the state of HTML entities in the exact form they were introduced, So `&` is preserved just as well as `&amp;`.

Can you test whether this fixes your specific issue and doesn't introduce a new problem?

@westonruter Thoughts on this approach?

westonruter:
This doesn't seem to work quite right. Consider this:

```php
<?php

$html = '<template type="amp-mustache">
    <a data-href="https://example.com/?foo={{foo}}&bar={{bar}}&amp;baz={{baz}}">Link</a>
</template></html>';

$doc = new \AmpProject\Dom\Document();
$doc->loadHTML( $html );

echo $doc->saveHTML();
```

Notice how there the `&` and `&amp;` is inside of an `data-href` attribute and not `href` here. The output here is:

```html
<!DOCTYPE html>
<html><head><meta charset="utf-8"></head><body><template type="amp-mustache">
    <a data-href="https://example.com/?foo={{foo}}&amp;bar={{bar}}&amp;amp;baz={{baz}}">Link</a>
</template></body></html>
```

Notice the `&amp;` is getting serialized as `&amp;amp;` which is not correct.

I'm going to dig deeper.

westonruter:
@schlessera I found that if the issue only occurs if updating an attribute value via `$attr->nodeValue`. It doesn't happen when updating an attribute value via `$el->setAttribute()`. Example code:

```php
$dom = new DOMDocument();
$dom->loadHTML( '<!DOCTYPE html><html data-test1="" data-test2=""></html>' );
$dom->documentElement->setAttribute( 'data-test1', 'foo&bar' );
$dom->documentElement->getAttributeNode( 'data-test2' )->nodeValue = 'foo&bar';
echo $dom->saveHTML();
```

Output:

```
Warning: main(): unterminated entity reference             bar in /in/UTRPQ on line 6
<!DOCTYPE html>
<html data-test1="foo&amp;bar" data-test2></html>
```

So I've updated the logic to do so. I've also made it only update the value if a replacement was actually made.

westonruter (approved):

schlessera:
LGTM. @buihoanghai Can you test whether this latest incarnation solves your particular issue?

PR #58, opened 2021-02-04
Add GHA workflow to sync test suite and create PR

schlessera:


schlessera:
Example of a PR created by this: https://github.com/ampproject/amp-toolbox-php/pull/66

.github/workflows/sync-spec-test-suite.yml
```diff
@@ -0,0 +1,72 @@
...
+          echo $BODY
+          BODY="${BODY//$'\n'/'%0A'}"
+          BODY="${BODY//$'`'/'\`'}"
+          echo -e "::set-output name=text::$BODY"
+
+      - name: Commit update
+        run: |
+          git add --all .
+          git commit -m "Sync spec test suite - $(date +'%Y-%m-%d')" || true
+          git push origin sync-spec-test-suite || true
```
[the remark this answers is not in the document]
schlessera: Yes, so that the action does not fail the checks. This would probably not be needed if I only run this conditionally as you suggest above.

.github/workflows/sync-local-fallback-files.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: composer install
+
+      - name: Create branch with committed changes
+        run: |
+          git config --global user.email "alain.schlesser@gmail.com"
+          git config --global user.name "Alain Schlesser"
+          git fetch
+          git branch -f sync-local-fallback-files origin/main
+          git checkout sync-local-fallback-files
+          git pull --no-rebase || true
```
westonruter, line 43: Is the `git pull` needed here? You just above created a new branch from `origin/master` after you `git fetch`, so then there should never be any new changes, right?

```suggestion
```

.github/workflows/sync-spec-test-suite.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: composer install
+
+      - name: Create branch as needed and pull latest changes
+        run: |
+          git config --global user.email "alain.schlesser@gmail.com"
+          git config --global user.name "Alain Schlesser"
+          git fetch
+          git branch -f sync-spec-test-suite origin/main
+          git checkout sync-spec-test-suite
+          git pull --no-rebase || true
```
westonruter, line 43: Ditto above.

```suggestion
```

.github/workflows/sync-local-fallback-files.yml
schlessera: The reason I have this in here is because:
- There might already be a branch with this name, which should be updated, as I don't want to script to create multiple PRs if we don't get to merge them right away: and
- On an existing PR, we will sometimes add manual commits to fix the tests that break with an automated update to make them pass again, so the script should only ever _add_ to an existing PR, and not _replace_ it and cause manual work to be lost.

westonruter: Thanks. Makes sense.

.github/workflows/sync-local-fallback-files.yml
```diff
@@ -0,0 +1,83 @@
...
+          git push origin sync-local-fallback-files
+
+      - name: Create pull request
+        if: steps.changes.outputs.changed == 1
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: "sync-local-fallback-files"
+          destination_branch: "main"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          pr_title: "Sync local fallback files"
```
westonruter, line 80: Should this have the date included as well? I see a date in the title here: https://github.com/ampproject/amp-toolbox-php/pull/57

.github/workflows/sync-spec-test-suite.yml
```diff
@@ -0,0 +1,83 @@
...
+          git push origin sync-spec-test-suite
+
+      - name: Create pull request
+        if: steps.changes.outputs.changed == 1
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: "sync-spec-test-suite"
+          destination_branch: "main"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          pr_title: "Sync spec test suite"
```
westonruter, line 80: Was a date provided originally here? I see it here: https://github.com/ampproject/amp-toolbox-php/pull/57

westonruter (approved):
PRs look good: https://github.com/ampproject/amp-toolbox-php/pull/66, https://github.com/ampproject/amp-toolbox-php/pull/57

I defer to @pierlon for final OK on the actions.

.github/workflows/sync-local-fallback-files.yml
```diff
@@ -0,0 +1,83 @@
...
+          git branch -f sync-local-fallback-files origin/main
+          git checkout sync-local-fallback-files
+          git pull --no-rebase || true
+
+      - name: Sync local fallback files
+        run: composer sync-fallback-files
+
+      - name: Check if there are changes
+        id: changes
+        uses: UnicornGlobal/has-changes-action@v1.0.11
```
pierlon, line 54: This action seems unnecessary as it's simply running `git status --porcelain` to determine if files have changed. This could be replaced with:

```suggestion
        run: echo ::set-output name=changed::$([[ -z $(git status --porcelain) ]] && echo "0" || echo "1")
```

.github/workflows/sync-spec-test-suite.yml
```diff
@@ -0,0 +1,83 @@
...
+          git branch -f sync-spec-test-suite origin/main
+          git checkout sync-spec-test-suite
+          git pull --no-rebase || true
+
+      - name: Sync spec test suite
+        run: composer sync-test-specs
+
+      - name: Check if there are changes
+        id: changes
+        uses: UnicornGlobal/has-changes-action@v1.0.11
```
pierlon, line 54: Same case as above:

```suggestion
        run: echo ::set-output name=changed::$([[ -z $(git status --porcelain) ]] && echo "0" || echo "1")
```

pierlon (changes requested):

.github/workflows/sync-local-fallback-files.yml
pierlon: Forgot to add:

This action adds a considerable amount of time to the overall execution time of the workflow. In [this check](https://github.com/ampproject/amp-toolbox-php/pull/58/checks?check_run_id=1844050880), you can see that it spends just about half of the time waiting for the Docker image to be built before it can continue.

schlessera:
This is ready to be used as-is. In the screenshot below one can see that the PRs are being updated and manual commits are preserved.

Image 2021-02-20 at 11 34 00 AM

The only thing that can be improved right now is that the PR description is not being updated, so if the files being changed by the PR are changing after a manual or follow-up commit, the PR description needs to be updated manually. However, given that one might also want to adapt the PR description manually and not want to lose those edits, I'm not sure how this could be done in a robust way, so let's stick with manual description updates only for now.

westonruter:
> The only thing that can be improved right now is that the PR description is not being updated, so if the files being changed by the PR are changing after a manual or follow-up commit, the PR description needs to be updated manually. However, given that one might also want to adapt the PR description manually and not want to lose those edits, I'm not sure how this could be done in a robust way, so let's stick with manual description updates only for now.

Maybe the description should just be omitted in that case? The PR description listing out the changed files is redundant with the Files tab of the PR anyway, right?

schlessera:
Given that:
- most of the times, a PR generated by this script will be merged either automatically or within a short time frame
- the description provides a good initial overview to decide whether the approval should be done in depth or not
- the description can be edited automatically, even after merging

... I'd say we keep the description around and (in case it ever becomes a problem) we look into making it update correctly.

schlessera:
I'm merging this for now, as I think it is "good enough" and should take any more time for now, and any future annoyances can be dealt with in subsequent PRs.

PR #79, opened 2021-02-23
Parse document correctly when AMP emoji is used

### LINKED ISSUE #75: Document parsing fails when HTML start tag contains ⚡
Hi there,

we just started using the AMP Optimizer library and found a bug. In our admittedly strange setup, we use a class binding on the html element to set some marker classes. These class binding gets removed when using the AMP Optimizer.

Input:
```html
<!doctype html>
<html ⚡ [class]="mystate.class" class="blue">
<head>
    <meta charset="utf-8">
    <title>My AMP Page</title>
    <link rel="canonical" href="self.html" />
    <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
    <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
    <script async src="https://cdn.ampproject.org/v0.js"></script>
    <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
    <style amp-custom>
        h1 {
            margin: 1rem;
        }
        .blue {
            background-color: blue;
        }
        .yellow {
            background-color: yellow;
        }
    </style>
</head>
<body>
<amp-state id="mystate">
    <script type="application/json">
        {
            "class": "blue"
        }
    </script>
</amp-state>
<button on="tap:AMP.setState({mystate: {class: 'yellow'}})">
    yellow
</button>
<button on="tap:AMP.setState({mystate: {class: 'blue'}})">
    blue
</button>
</body>
</html>
```

Output (the relevant part):
```html
<!DOCTYPE html>
<html ⚡ class="blue" i-amphtml-layout="" i-amphtml-no-boilerplate="" transformed="self;v=1">
<head><meta charset="utf-8">
```
So after the optimization, clicking on the buttons has no effect.

We use the Optimizer with the default configuration (nothing special).

schlessera:
Fixes #75

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -66,7 +66,7 @@ final class Document extends DOMDocument
     /**
      * Associative array of known options and their respective default value.
      *
-     * @var array<string>
+     * @var array
      */
     const KNOWN_OPTIONS = [
         self::OPTION_AMP_BIND_SYNTAX => self::AMP_BIND_SYNTAX_AUTO,
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -619,6 +619,16 @@ final class Document extends DOMDocument
     }
 
     /**
+     * Get the current options of the Document instance.
+     *
+     * @return array
+     */
+    public function getOptions()
+    {
+        return $this->options;
+    }
+
+    /**
      * Add the required utf-8 meta charset tag if it is still missing.
      */
     private function insertMissingCharset()
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -37,6 +37,37 @@ use DOMXPath;
  */
 final class Document extends DOMDocument
 {
+    /**
+     * Option to configure the preferred amp-bind syntax.
+     *
+     * @var string
+     */
+    const OPTION_AMP_BIND_SYNTAX = 'amp_bind_syntax';
+
+    /**
+     * Option to provide the encoding of the document.
+     *
+     * @var string
+     */
+    const OPTION_ENCODING = 'encoding';
+
+    /**
+     * Option to provide additional libxml flags to configure parsing of the document.
+     *
+     * @var string
+     */
+    const OPTION_LIBXML_FLAGS = 'libxml_flags';
+
+    /**
+     * Associative array of known options and their respective default value.
+     *
+     * @var array<string>
+     */
+    const KNOWN_OPTIONS = [
+        self::OPTION_AMP_BIND_SYNTAX => 'auto',
+        self::OPTION_ENCODING        => null,
+        self::OPTION_LIBXML_FLAGS    => 0,
+    ];
 
     /**
      * AMP requires the HTML markup to be encoded in UTF-8.
@@ -334,15 +365,23 @@ final class Document extends DOMDocument
     /**
      * Named constructor to provide convenient way of transforming HTML into DOM.
      *
-     * @param string $html     HTML to turn into a DOM.
-     * @param string $encoding Optional. Encoding of the provided HTML string.
+     * @param string       $html    HTML to turn into a DOM.
+     * @param array|string $options Optional. Array of options to configure the document. Used as encoding if a string
+     *                              is passed. Defaults to an empty array.
      * @return Document|false DOM generated from provided HTML, or false if the transformation failed.
      */
-    public static function fromHtml($html, $encoding = null)
+    public static function fromHtml($html, $options = [])
     {
+        // Assume options are the encoding if a string is passed, for BC reasons.
+        if (is_string($options)) {
+            $options = [self::OPTION_ENCODING => $options];
+        }
+
+        $encoding = isset($options[self::OPTION_ENCODING]) ? $options[self::OPTION_ENCODING] : null;
+
         $dom = new self('', $encoding);
 
-        if (! $dom->loadHTML($html)) {
+        if (! $dom->loadHTML($html, $options)) {
             return false;
         }
 
@@ -354,15 +393,23 @@ final class Document extends DOMDocument
      *
      * The difference to Document::fromHtml() is that fragments are not normalized as to their structure.
      *
-     * @param string $html     HTML to turn into a DOM.
-     * @param string $encoding Optional. Encoding of the provided HTML string.
+     * @param string       $html    HTML to turn into a DOM.
+     * @param array|string $options Optional. Array of options to configure the document. Used as encoding if a string
+     *                              is passed. Defaults to an empty array.
      * @return Document|false DOM generated from provided HTML, or false if the transformation failed.
      */
-    public static function fromHtmlFragment($html, $encoding = null)
+    public static function fromHtmlFragment($html, $options = [])
     {
+        // Assume options are the encoding if a string is passed, for BC reasons.
+        if (is_string($options)) {
+            $options = [self::OPTION_ENCODING => $options];
+        }
+
+        $encoding = isset($options[self::OPTION_ENCODING]) ? $options[self::OPTION_ENCODING] : null;
+
         $dom = new self('', $encoding);
 
-        if (! $dom->loadHTMLFragment($html)) {
+        if (! $dom->loadHTMLFragment($html, $options)) {
             return false;
         }
 
@@ -425,11 +472,12 @@ final class Document extends DOMDocument
      *
      * @link  https://php.net/manual/domdocument.loadhtml.php
      *
-     * @param string     $source  The HTML string.
-     * @param int|string $options Optional. Specify additional Libxml parameters.
+     * @param string           $source  The HTML string.
+     * @param array|int|string $options Optional. Array of options to configure the document. Used as additional Libxml
+     *                                  parameters if an int or string is passed. Defaults to an empty array.
      * @return bool true on success or false on failure.
      */
-    public function loadHTML($source, $options = 0)
+    public function loadHTML($source, $options = [])
     {
         $source  = $this->normalizeDocumentStructure($source);
         $success = $this->loadHTMLFragment($source, $options);
@@ -451,12 +499,23 @@ final class Document extends DOMDocument
     /**
      * Load a HTML fragment from a string.
      *
-     * @param string     $source  The HTML fragment string.
-     * @param int|string $options Optional. Specify additional Libxml parameters.
+     * @param string           $source  The HTML fragment string.
+     * @param array|int|string $options Optional. Array of options to configure the document. Used as additional Libxml
+     *                                  parameters if an int or string is passed. Defaults to an empty array.
      * @return bool true on success or false on failure.
      */
-    public function loadHTMLFragment($source, $options = 0)
+    public function loadHTMLFragment($source, $options = [])
     {
+        // Assume options are the additional libxml flags if a string or int is passed, for BC reasons.
+        if (is_string($options)) {
+            $options = (int) $options;
+        }
+        if (is_int($options)) {
+            $options = [self::OPTION_LIBXML_FLAGS => $options];
+        }
+
+        $options = array_merge(self::KNOWN_OPTIONS, $options);
+
         $this->reset();
 
         $source = $this->convertAmpEmojiAttribute($source);
@@ -483,7 +542,7 @@ final class Document extends DOMDocument
 
         $libxml_previous_state = libxml_use_internal_errors(true);
 
-        $options |= LIBXML_COMPACT;
+        $options[self::OPTION_LIBXML_FLAGS] |= LIBXML_COMPACT;
 
         /*
          * LIBXML_HTML_NODEFDTD is only available for libxml 2.7.8+.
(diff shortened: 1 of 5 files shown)


src/Dom/Document.php
```diff
@@ -1099,6 +1110,38 @@ private function convertAmpBindAttributes($html)
...
+            self::AMP_BIND_DATA_ATTR_PREFIX,
+            implode('|', array_unique($this->convertedAmpBindAttributes))
+        );
+        $replacement = '[$1]';
+        $limit       = count($this->convertedAmpBindAttributes);
+
+        $restored = preg_replace($pattern, $replacement, $html, $limit);
+
+        return (null !== $restored) ? $restored : $html;
+    }
```
westonruter, line 1143: I still don't believe that _restoring_ is the right approach here since attributes can be added/removed after the DOM has been parsed and before it has been serialized, resulting in a incorrect restoration as the counts will change: https://github.com/ampproject/amp-toolbox-php/commit/bb607f85ae3e277d67627d03b9e0ea80f6ab2476#commitcomment-47328593

As mentioned in https://github.com/ampproject/amp-toolbox-php/commit/bb607f85ae3e277d67627d03b9e0ea80f6ab2476#commitcomment-47328668:

> Therefore, I think the only reasonable path is to introduce a `useBracketedAmpBindSyntax` flag which controls whether _all_ attributes are serialized using the bracketed syntax or the `data-amp-bind-` syntax. By default it would be `false`, but in unit tests you could make it `true`.

westonruter (changes requested):

src/Dom/Document.php
schlessera: So, what I did now is to provide a way to pass options around the document (with BC fallback that mimic the built-in `DOMDocument` do its syntax still works) and I added an option (`Document::OPTION_AMP_BIND_SYNTAX`) to configure how the conversion should be handled for `amp-bind` attributes. There are three available options:
- `Document::AMP_BIND_SYNTAX_AUTO` => try a best guess to keep the individual attributes in the syntax they were in originally.
- `Document::AMP_BIND_SYNTAX_DATA_ATTRIBUTE` => convert all `amp-bind` syntax into the `data-amp-bind-*` syntax.
- `Document::AMP_BIND_SYNTAX_SQUARE_BRACKETS` => convert all `amp-bind` attributes into the `[*]` syntax.

We can then set the option to `Document::AMP_BIND_SYNTAX_DATA_ATTRIBUTE` in the WordPress plugin to keep the current behavior for WP unchanged.

The default is `Document::AMP_BIND_SYNTAX_AUTO`, as I think this is the expected behavior, as third-party users of the library would not even be aware the `DOMDocument` code is broken and requires such a behind-the-scenes conversion in the first place.

src/Dom/Document.php
```diff
@@ -327,15 +379,23 @@ public function __construct($version = '', $encoding = null)
...
+    public static function fromHtml($html, $options = [])
     {
+        // Assume options are the encoding if a string is passed, for BC reasons.
+        if (is_string($options)) {
+            $options = [self::OPTION_ENCODING => $options];
+        }
+
+        $encoding = isset($options[self::OPTION_ENCODING]) ? $options[self::OPTION_ENCODING] : null;
+
         $dom = new self('', $encoding);
```
westonruter, line 396: Alternatively:

```suggestion
        $options = array_merge( self::KNOWN_OPTIONS, $options );

        $dom = new self('', $options[self::OPTION_ENCODING]);
```

src/Dom/Document.php
```diff
@@ -347,15 +407,23 @@ public static function fromHtml($html, $encoding = null)
...
+    public static function fromHtmlFragment($html, $options = [])
     {
+        // Assume options are the encoding if a string is passed, for BC reasons.
+        if (is_string($options)) {
+            $options = [self::OPTION_ENCODING => $options];
+        }
+
+        $encoding = isset($options[self::OPTION_ENCODING]) ? $options[self::OPTION_ENCODING] : null;
+
         $dom = new self('', $encoding);
```
westonruter, line 424:
```suggestion
        $options = array_merge( self::KNOWN_OPTIONS, $options );

        $dom = new self('', $options[self::OPTION_ENCODING]);
```

src/Dom/Document.php
westonruter: But I see this merge is already being done below.

src/Dom/Document.php
```diff
@@ -1099,6 +1183,54 @@ private function convertAmpBindAttributes($html)
...
+        $pattern = $this->options[self::OPTION_AMP_BIND_SYNTAX] === self::AMP_BIND_SYNTAX_AUTO
+            ? sprintf(
+                '#%s(%s)#i',
+                self::AMP_BIND_DATA_ATTR_PREFIX,
+                implode('|', array_unique($this->convertedAmpBindAttributes))
+            )
+            : sprintf(
+                '#%s([a-z0-9-_]+)#i', // TODO: Check what exact regex makes sense here.
+                self::AMP_BIND_DATA_ATTR_PREFIX
+            );
```
westonruter, line 1225: Don't the patterns here run the risk of unintentional conversion? For example, if I write a post about amp-bind and I say:

> You have two options: use the `[text]` attribute or the `data-amp-bind-text` attribute.

At present this will result in the unexpected:

> You have two options: use the `[text]` attribute or the `[text]` attribute.

In order to safely convert back to the bracketed syntax, we'd need to use the same logic in `convertAmpBindAttributes` which uses the `AMP_BIND_START_TAGS_PATTERN` regex to first match all tags which contain an attribute in question, and then in the replacement callback do the actual replacements.

src/Dom/Document.php
westonruter: Ditto above, but nevermind per below.

westonruter (changes requested):

src/Dom/Document.php
schlessera: Yes, I wanted to avoid multiple merges.

src/Dom/Document.php
schlessera: Yes, indeed, this needs to become way more robust.

schlessera: I think this should now work as expected (see added tests with latest commits).

The regex I ended up with looks overly complex, so happy to hear feedback if I'm missing something obvious.

src/Dom/Document.php
```diff
@@ -1082,7 +1201,93 @@ private function convertAmpBindAttributes($html)
...
+            while (preg_match(self::AMP_BIND_DATA_ATTRIBUTE_ATTR_PATTERN, substr($oldAttrs, $offset), $attrMatches)) {
+                $offset += strlen($attrMatches[0]);
+
+                $attrName = substr($attrMatches['name'], strlen(self::AMP_BIND_DATA_ATTR_PREFIX));
+                if (
+                    $this->options[self::OPTION_AMP_BIND_SYNTAX] === self::AMP_BIND_SYNTAX_SQUARE_BRACKETS
+                    ||
+                    in_array($attrName, $this->convertedAmpBindAttributes, true)
+                ) {
+                    $newAttrs .= " [{$attrName}]";
```
westonruter, line 1272: Minor point: the replacement includes a single `' '` space. The regex, however, matches any repeated sequence of whitespace. Granted the inter-attribute whitespace at this point would have been normalize by the DOM serialization, but it's just something I wanted to point out. Probably no action required.

src/Dom/Document.php
```diff
@@ -1082,7 +1201,93 @@ private function convertAmpBindAttributes($html)
...
+                    if (isset($attrMatches['value'])) {
```
westonruter, line 1273: Per note above, the value of an AMP-bind attribute should actually always be provided present. A bind attribute by definition has to have a value to control the state which should manage the bound attribute.

src/Dom/Document.php
```diff
@@ -1397,14 +1602,49 @@ private function convertAmpEmojiAttribute($source)
...
+                $source = preg_replace(
+                    self::AMP_EMOJI_ATTRIBUTE_PATTERN,
+                    '<html ' . implode(' ', $attributes) . '>',
+                    $source
+                );
+                break;
+            }
+        }
+
+        return $source;
```
westonruter, line 1647: This could be made more efficient by adapting into a `preg_replace_callback` where the following code is put inside the callback. In that way, there would not be a need to do an additional `preg_replace()` on the `$source`. It could be combined with teh `preg_match()` call above as well. Something like this I believe:

```php
    private function convertAmpEmojiAttribute($source)
    {
        $this->usedAmpEmoji = '';

        return preg_replace_callback(
            self::AMP_EMOJI_ATTRIBUTE_PATTERN,
            static function ( $matches ) {
                // Split into individual attributes.
                $attributes = array_map(
                    'trim',
                    array_filter(
                        preg_split(
                            '#(\s+[^"\'\s=]+(?:=(?:"[^"]+"|\'[^\']+\'|[^"\'\s]+))?)#',
                            $matches[2],
                            -1,
                            PREG_SPLIT_DELIM_CAPTURE
                        )
                    )
                );

                foreach ($attributes as $index => $attribute) {
                    $attributeMatches = [];
                    if (
                        preg_match(
                            '/^(' . Attribute::AMP_EMOJI_ALT . '|' . Attribute::AMP_EMOJI . ')(4(?:ads|email))?$/i',
                            $attribute,
                            $attributeMatches
                        )
                    ) {
                        $this->usedAmpEmoji = $attributeMatches[1];
                        $variant            = ! empty($attributeMatches[2]) ? $attributeMatches[2] : '';
                        $attributes[$index] = self::EMOJI_AMP_ATTRIBUTE_PLACEHOLDER . "=\"{$variant}\"";
                        break;
                    }
                }

                return '<html ' . implode(' ', $attributes) . '>';
            },
            $source,
            1
        );
    }
```

Note also the added limit of 1.

westonruter (changes requested):

src/Dom/Document.php
schlessera: I adapted to code to use a single `preg_replace_callback()`. Your code was already pretty close, but it needed a bit more fiddling with the regex pattern(s).

Just to be sure, I profiled the difference across the unit tests we have that cover emojis, and this is the result:

Comparison from Untitled to Untitled - Blackfire - Google Chrome 2021-03-15 at 11 22 19 AM

So it's indeed a clear improvement overall.

src/Dom/Document.php
```diff
@@ -37,6 +37,42 @@
...
 final class Document extends DOMDocument
 {
+    /**
+     * Option to configure the preferred amp-bind syntax.
+     *
+     * @var string
+     */
+    const OPTION_AMP_BIND_SYNTAX = 'amp_bind_syntax';
+
+    // Different flags for the amp-bind syntax option.
```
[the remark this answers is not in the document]
schlessera: While doing so, I decided that it's time to refactor the `Document` class, as it is becoming huge and cumbersome. For now I just extracted some constants into two interfaces `Document\Option` and `Document\Encoding` to not overload this PR. I'll create issues for additional refactoring and deal with them in separate PRs.

src/Dom/Document.php
schlessera: I'm aware of the discrepancy but opted to keep it for now for the added robustness of accepting more than one whitespace.

src/Dom/Document.php
schlessera: I'm now falling back to an empty value (`=""`) instead of generating a boolean attribute.

src/Dom/Document.php
```diff
@@ -87,27 +123,62 @@ final class Document extends DOMDocument
     const AMP_BIND_DATA_ATTR_PREFIX = 'data-amp-bind-';
 
     /**
-     * Pattern for HTML attribute accounting for binding attr name, boolean attribute, single/double-quoted attribute
```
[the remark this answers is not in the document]
schlessera: Yes, binding attributes would always need a value. However, I opted to keep the matching intact so as to ensure even bad markup is correctly parsed. I do now enforce adding the value in the conversion below, though.

schlessera:
While looking into the few lines Codecov marked as not being tested, I noticed bugs in the encoding code and fixed them in here as well (and added a few tests for them too).

src/Dom/Document.php
```diff
@@ -1313,21 +1248,30 @@ public function restoreAmpBindAttributes($html)
...
-            $this->originalEncoding = mb_detect_encoding($source, self::ENCODING_DETECTION_ORDER, true);
+        if (Encoding::UNKNOWN === $this->originalEncoding && function_exists('mb_detect_encoding')) {
+            $this->originalEncoding = mb_detect_encoding($source, Encoding::DETECTION_ORDER, true);
         }
 
         // Guessing the encoding seems to have failed, so we assume UTF-8 instead.
+        // In my testing, this was not possible as long as one ISO-8859-x is in the detection order.
         if (empty($this->originalEncoding)) {
-            $this->originalEncoding = self::AMP_ENCODING;
+            $this->originalEncoding = Encoding::AMP;
```
westonruter, line 1258: If it's not possible, might as well exclude it from code coverage reporting:

```suggestion
            $this->originalEncoding = Encoding::AMP; // @codeCoverageIgnore
```

westonruter (approved):

PR #86, opened 2021-02-27
Add support for a data-hero-candidate attribute

schlessera:
This PR changes the `PreloadHeroImages` transformer to use the following logic:
1. Use all elements with `data-hero` attribute up to the max of 2 (throw error when past max).
2. If no `data-hero` attribute is found use all elements with `data-hero-candidate` attribute up to the max of 2 (no errors thrown).
3. If neither `data-hero` nor `data-hero-candidate` attribute is found, use first element that qualifies as "hero image fallback".

### DIFF
--- a/src/Optimizer/Transformer/PreloadHeroImage.php
+++ b/src/Optimizer/Transformer/PreloadHeroImage.php
@@ -215,13 +215,15 @@ final class PreloadHeroImage implements Transformer
                     $heroImageCandidates[] = $heroImageCandidate;
                 } elseif (count($heroImageFallbacks) < self::DATA_HERO_MAX) {
                     $heroImageFallback = $this->detectPossibleHeroImageFallbacks($node);
-                    if ($heroImageFallback) {
+                    if (
+                        $heroImageFallback &&
                         // Ensure we don't flag the same image twice. This can happen for placeholder images, which are
                         // flagged on their own and as their parent's placeholder.
-                        if (
+                        (
                             ! $previousHeroImageFallback
                             || $heroImageFallback->getAmpImg() !== $previousHeroImageFallback->getAmpImg()
-                        ) {
+                        )
+                    ) {
                             $heroImageFallbacks[]      = $heroImageFallback;
                             $previousHeroImageFallback = $heroImageFallback;
                         }
--- a/src/Optimizer/Transformer/PreloadHeroImage.php
+++ b/src/Optimizer/Transformer/PreloadHeroImage.php
@@ -215,18 +215,18 @@ final class PreloadHeroImage implements Transformer
                     $heroImageCandidates[] = $heroImageCandidate;
                 } elseif (count($heroImageFallbacks) < self::DATA_HERO_MAX) {
                     $heroImageFallback = $this->detectPossibleHeroImageFallbacks($node);
+
+                    // Ensure we don't flag the same image twice. This can happen for placeholder images, which are
+                    // flagged on their own and as their parent's placeholder.
                     if (
-                        $heroImageFallback &&
-                        // Ensure we don't flag the same image twice. This can happen for placeholder images, which are
-                        // flagged on their own and as their parent's placeholder.
-                        (
+                        $heroImageFallback
+                        && (
                             ! $previousHeroImageFallback
                             || $heroImageFallback->getAmpImg() !== $previousHeroImageFallback->getAmpImg()
                         )
                     ) {
-                            $heroImageFallbacks[]      = $heroImageFallback;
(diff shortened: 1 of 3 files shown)


src/Optimizer/Transformer/PreloadHeroImage.php
```diff
@@ -194,24 +194,41 @@ public function transform(Document $document, ErrorCollection $errors)
...
+                    $heroImageCandidates[] = $heroImageCandidate;
+                } elseif (count($heroImageFallbacks) < self::DATA_HERO_MAX) {
+                    $heroImageFallback = $this->detectPossibleHeroImageFallbacks($node);
+                    if ($heroImageFallback) {
+                        // Ensure we don't flag the same image twice. This can happen for placeholder images, which are
+                        // flagged on their own and as their parent's placeholder.
+                        if (
+                            ! $previousHeroImageFallback
+                            || $heroImageFallback->getAmpImg() !== $previousHeroImageFallback->getAmpImg()
+                        ) {
```
westonruter, line 224: Should these two conditions be combined? 

```suggestion
                    if (
                        $heroImageFallback &&
                        // Ensure we don't flag the same image twice. This can happen for placeholder images, which are
                        // flagged on their own and as their parent's placeholder.
                        (
                            ! $previousHeroImageFallback
                            || $heroImageFallback->getAmpImg() !== $previousHeroImageFallback->getAmpImg()
                        )
                    ) {
```

tests/Optimizer/Transformer/PreloadHeroImageTest.php
```diff
@@ -205,6 +205,55 @@ public function dataTransform()
...
+                ),
+                $output(
+                    '<amp-img width="500" height="400" src="/foo.png"></amp-img>'
+                    . '<amp-img data-hero-candidate width="500" height="400" src="/hero1.png"></amp-img>'
+                    . '<amp-img data-hero-candidate width="500" height="400" src="/hero2.png"></amp-img>'
+                    . '<amp-img data-hero-candidate width="500" height="400" src="/hero3.png"></amp-img>'
+                    . '<amp-img data-hero-candidate width="500" height="400" src="/hero4.png"></amp-img>'
+                    . '<amp-img data-hero width="500" height="400" src="/hero5.png" i-amphtml-ssr><img class="i-amphtml-fill-content i-amphtml-replaced-content" decoding="async" src="/hero5.png"></amp-img>'
+                ),
+            ],
```
westonruter, line 256: Can tests be added for fallbacks as well as testing the placeholder fallback image deduplication logic?

src/Optimizer/Transformer/PreloadHeroImage.php
```diff
@@ -220,29 +237,33 @@ private function findHeroImages(Document $document)
...
         }
 
-        // Fall back to auto-detected hero image if available.
-        if ($heroImageCandidate) {
-            return [$heroImageCandidate];
+        while (count($heroImages) < self::DATA_HERO_MAX && count($heroImageCandidates) > 0) {
+            $heroImages[] = array_shift($heroImageCandidates);
+        }
+
+        if (count($heroImages) < 1 && count($heroImageFallbacks) > 0) {
```
westonruter, line 248: If the hero image candidates don't make up the maximum hero images allowed, should not the first 1-2 fallback images also backfill the hero images?

```suggestion
        while (count($heroImages) < self::DATA_HERO_MAX && count($heroImageFallbacks) > 0) {
```

westonruter (changes requested):

src/Optimizer/Transformer/PreloadHeroImage.php
schlessera: I had this in the beginning but then decided against it, as it can lead to unexpected problems.

The fallback images are basically collected from the first big images based on markup order. If you use something like we have with `DetermineHeroImages` to be smarter about what to pick, filling it up with a fallback image could mean the fallback image was something specifically skipped in `DetermineHeroImages` but still earlier in markup order than the actual candidate.

I'd prefer "fallback" images to be a fallback in case nothing at all had been marked, and as soon as _something_ was marked, we assume the marking was correct and no fallback are needed or will make things worse.

tests/Optimizer/Transformer/PreloadHeroImageTest.php
schlessera: Those are already tested. The fallbacks are tested in the first 2 tests (`detects regular image as hero image` & `skips tiny images`), for example. The placeholders are tested with the tests that are using placeholders, like `fetches placeholders for animations`, as these would otherwise have the SSR'ed `` element twice (which was what detected the bug in the first place).

PR #99, opened 2021-03-16
Fix broken encoding when using document fragments

### LINKED ISSUE #28: Encoding issue when loading HTML fragment
The following code produces broken encoding within `Dom\Document`:

```php
$html     = '<div style="Iñtërnâtiônàlizætiøn"></div>';
$document = Document::fromHtmlFragment($html);
var_dump($document)->saveHtml();

// Output:
// string(101) "<html><head></head><body><div style="IÃ±tÃ«rnÃ¢tiÃ´nÃ lizÃ¦tiÃ¸n"></div></body></html>\n"
```

However, the following code seems to work just fine (note the added `<head>` tag):

```php
$html     = '<head></head><div style="Iñtërnâtiônàlizætiøn"></div>';
$document = Document::fromHtmlFragment($html);
var_dump($document)->saveHtml();

// Output:
// string(87) "<html><head></head><body><div style="Iñtërnâtiônàlizætiøn"></div></body></html>\n"
```

This needs investigation. My first assumption is that we do some normalization that requires the `<head>` tag and silently fails when it is missing.

schlessera:
This ensures that the `http-equiv` charset meta tag needed for making the `DOMDocument` work properly with UTF-8 encoding is always added, even when all or parts of the HTML document structure are missing.

The current implementation does multiple regex calls, with the assumption that this optimizes for large documents where a `<head>` tag would normally always be present. So for any full real-world HTML document, only the first regex would ever be used. For actual document fragments, this assumes they would be small anyway, making multiple regex traversals cheap in these cases.

These multiple regex calls could be combined into one, but this would likely make the most common case much worse in terms of performance. Thoughts on that, @westonruter ?

Fixes #28

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -177,8 +177,6 @@ final class Document extends DOMDocument
 
     const HTML_GET_HEAD_OPENING_TAG_PATTERN     = '/(?><!--.*?-->\s*)*<head(?>\s+[^>]*)?>/is';
     const HTML_GET_HEAD_OPENING_TAG_REPLACEMENT = '$0' . self::HTTP_EQUIV_META_TAG;
-    const HTML_GET_BODY_OPENING_TAG_PATTERN     = '/(?><!--.*?-->\s*)*<body(?>\s+[^>]*)?>/is';
-    const HTML_GET_BODY_OPENING_TAG_REPLACEMENT = '<head>' . self::HTTP_EQUIV_META_TAG . '</head>$0';
     const HTML_GET_HTML_OPENING_TAG_PATTERN     = '/(?><!--.*?-->\s*)*<html(?>\s+[^>]*)?>/is';
     const HTML_GET_HTML_OPENING_TAG_REPLACEMENT = '$0<head>' . self::HTTP_EQUIV_META_TAG . '</head>';
     const HTML_GET_HTTP_EQUIV_TAG_PATTERN       = '#<meta http-equiv=([\'"])content-type\1 '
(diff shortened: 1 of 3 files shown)


src/Dom/Document.php
```diff
@@ -2044,4 +2049,67 @@ public function enforceCssMaxByteCount($maxByteCount = AMP::MAX_CSS_BYTE_COUNT)
...
+                $html,
+                1,
+                $count
+            );
+        }
+
+        // Finally, we just prepend the head with the required http-equiv charset.
+        if ($count < 1) {
+            $html = '<head>' . self::HTTP_EQUIV_META_TAG . '</head>' . $html;
+        }
```
westonruter, line 2111: All of these may be a bit excessive. 

Consider this markup:

```php
$dom = new DOMDocument();
$dom->loadHTML( '<meta id="injected-meta-charset" http-equiv="content-type" content="text/html; charset=utf-8"><p>¡Fragment!</p>' );
echo $dom->saveHTML();
```

The output is:

```html
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><head><meta id="injected-meta-charset" http-equiv="content-type" content="text/html; charset=utf-8"></head><body><p>¡Fragment!</p></body></html>
```

So the parser automatically puts the `meta` charset in a `head` that it constructs dynamically.

I think all that is needed is if there is no `<html>` in the document, then just prepend the `meta` tag and the parser can fix it up.

schlessera: The `Dom\Document` has different default behavior than `DOMDocument`. Amongst others, it uses `LIBXML_HTML_NODEFDTD`, which prevents the default doctype (and as a side-effect, the `<head>` node) to be added automatically, as that was needed to allow for comments before the doctype (see https://github.com/ampproject/amp-wp/issues/4104#issuecomment-574216012).
I'll have to test how adding a meta tag without a `<head>` node behaves now.

schlessera: Hmm, the HTML one was in there twice, and the doctype one was not actually implemented... 🤦

schlessera: Reduced the needed code to what the tests showed was the minimum required. Fixed via https://github.com/ampproject/amp-toolbox-php/pull/99/commits/4758e217229fb333a029cdc238c07cf2d459466c

westonruter (approved):

PR #100, opened 2021-03-16
Add validator spec via code generation

schlessera:
This PR adds an object hierarchy that represents the validator spec.

It downloads the JSON version of the validator spec data and uses code generation to build the object hierarchy as a result. The code generation can be manually retriggered for a refresh via the following command:
```bash
php bin/generate-validator-spec.php
```

The generated validator spec is found under `AmpProject\Validator\Spec` (in the `src/Validator/Spec` subfolder), whereas the code generation is found under (the `--dev` only) `AmpProject\Tooling\Validator\SpecGenerator` (in the `bin/Validator/SpecGenerator` subfolder).


westonruter:
LGTM ✅ 

Just kidding. As you noted, I'll have to digest this. Maybe we could step through it next week so I can get a high-level overview before getting into the details?

westonruter (approved):
To do full post-merge review as discussed.

bin/generate-validator-spec.php
```diff
@@ -37,14 +38,31 @@ function recursivelyRemoveDirectory($directory)
...
     $specGenerator->generate(
-        json_decode(file_get_contents('https://cdn.ampproject.org/v0/validator.json'), true),
+        json_decode($json, true),
         'AmpProject\Validator',
         $destination
     );
+    echo "\n";
+
+    echo "Done!\n";
+
```
westonruter, line 64:
```suggestion
```

westonruter:
> To do full post-merge review as discussed.

See #221.

PR #119, opened 2021-04-01
Add RewriteAmpUrls transformer

### LINKED ISSUE #20: Extend Optimizer with transformer for module/nomodule scripts (ESM)
## Feature description

See https://github.com/ampproject/amp-toolbox/pull/722.

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
This PR adds the `RewriteAmpUrls` transformer that enables the following features:
- ES modules support (enabled by default)
- LTS support
- Custom prefix support
- Custom runtime version support
- Custom `geo-api` URL

This PR also moves some code from `AMP_DOM_Utils` into the library to enable the library to use `copyAttributes()`.

This PR also includes changes to the `ReorderHead` transformer to make it work properly with module/nomodule script combinations.

Fixes #20


tests/spec/transformers/valid/RewriteAmpUrls/adds_esm/input.html
```diff
@@ -1,3 +1,8 @@
+<!--
+{
+  "esmModulesEnabled": true
```
westonruter, line 3: If `esmModulesEnabled`, then why aren't aren't the `mjs` scripts and `nomodule` showing up here? May be same thing as I saw in https://github.com/ampproject/amp-wp/pull/6037#pullrequestreview-627637047.

schlessera: This is the input file, so the non-optimized version. The output file does contain the expected files: https://github.com/ampproject/amp-toolbox-php/blob/main/tests/spec/transformers/valid/RewriteAmpUrls/adds_esm/expected_output.html

westonruter (changes requested):

src/Dom/Element.php
```diff
@@ -77,6 +93,170 @@ public function setAttribute($name, $value)
...
+
+        return $attributes;
+    }
+
+    /**
+     * Add one or more HTML element attributes to this element.
+     *
+     * @param string[] $attributes One or more attributes for the node's HTML element.
+     */
+    public function addAttributes($attributes)
```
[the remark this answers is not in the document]
schlessera: In my head, this is about adding one or more attribute nodes to the collection of attribute nodes of that element.

But `setAttributes()` is more consistent with the existing API, so I'll rename.

src/Optimizer/Configuration/PreloadHeroImageConfiguration.php
```diff
@@ -48,8 +52,8 @@ final class PreloadHeroImageConfiguration extends BaseTransformerConfiguration
     protected function getAllowedKeys()
     {
         return [
-            self::PRELOAD_HERO_IMAGE            => true,
             self::INLINE_STYLE_BACKUP_ATTRIBUTE => '',
+            self::PRELOAD_HERO_IMAGE            => true,
```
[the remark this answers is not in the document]
schlessera: No, I'd change that in a different PR and in a different way.

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -0,0 +1,375 @@
...
+    private function adaptForSelfHosting(Document $document, $host, $errors)
+    {
+        // runtime-host and amp-geo-api meta tags should appear before the first script.
+        if (
+            ! $this->usesAmpCacheUrl($host)
+            &&
+            ! $this->configuration->get(RewriteAmpUrlsConfiguration::LTS)
+        ) {
+            try {
+                $urlParts = parse_url($host);
```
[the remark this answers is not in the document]
schlessera: Well, we already have a `Url` class in this library, so we can just extend it to do what we need here: https://github.com/ampproject/amp-toolbox-php/blob/main/src/Url.php

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -0,0 +1,375 @@
...
+
+        while (
+            $referenceNode !== null
+            &&
+            $referenceNode->nextSibling instanceof Element
+            &&
+            $referenceNode->nextSibling->tagName === Tag::LINK
+        ) {
+            $referenceNode = $referenceNode->nextSibling;
+        }
```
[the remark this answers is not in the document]
schlessera: I thought this would fix the test failures in the plugin, but it's not enough. The only proper fix is to fix #125 and put the `ReorderHead` transformer last again. Everything else is just random juggling at this point anyway.

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -0,0 +1,375 @@
...
+        $preloadNode->setAttribute(Attribute::AS_, $type);
+
+        return $preloadNode;
+    }
+
+    /**
+     * Add meta tags as needed to adapt for self-hosting the AMP runtime.
+     *
+     * @param Document        $document Document to add the meta tags to.
+     * @param string          $host     Host URL to use.
```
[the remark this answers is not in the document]
westonruter: Well, since it says "Host _URL_" then this is fine. It's just a slightly different meaning of "host".

src/Optimizer/Transformer/RewriteAmpUrls.php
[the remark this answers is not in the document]
westonruter: Hah! I should have known.

src/Optimizer/Transformer/RewriteAmpUrls.php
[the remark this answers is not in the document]
westonruter: I believe this is fixed now by #130.

src/Optimizer/Transformer/RewriteAmpUrls.php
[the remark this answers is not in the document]
schlessera: I fleshed out the `Url` class some more. It has a bit of a weird logic to take the multiple existing use cases into account but it should work now for what is required.

src/Url.php
```diff
@@ -2,70 +2,298 @@
...
+        return $url;
+    }
+
+    /**
+     * Magic getter to return the individual parts.
+     *
+     * @param string $name Name of the part to return.
+     * @return string|null Part string or null if it was not found during parsing.
+     */
+    public function __get($name)
```
westonruter, line 289: Eventually we should implement `__set()` as well, since it would be very useful to be able to change out just the `host`, when rewriting a URL to use the AMP Cache for example. 

Either this, or make the member variables `public`.

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -268,9 +269,14 @@ private function adaptForSelfHosting(Document $document, $host, $errors)
...
         ) {
             try {
-                $urlParts = parse_url($host);
-                $origin   = "{$urlParts['scheme']}://{$urlParts['host']}";
-                $this->addMeta($document, 'runtime-host', $origin);
+                $url = new Url($host);
+
+                if (!empty($url->scheme) && !empty($url->host)) {
+                    $origin = "{$url->scheme}://{$url->host}";
+                    $this->addMeta($document, 'runtime-host', $origin);
```
westonruter, line 276: What about scheme-relative URLs? In that case, `$url->scheme` would be empty and the `$origin` would be `"//{$url->host}"`. I guess that's not supported, and we'd want an explicit HTTPS URL anyway.

src/Url.php
```diff
@@ -2,70 +2,298 @@
...
+        if ($this->scheme === 'data') {
             return false;
         }
 
-        $userpass = $user;
+        // @TODO: This probably needs additional logic.
+
+        $parts = array_merge(self::URL_DEFAULT_PARTS, $this->toArray(true));
+
+        return (bool)filter_var(self::toString($parts), FILTER_VALIDATE_URL);
```
westonruter, line 210: We'll have to add `filter` to the required extensions.

src/Url.php
```diff
@@ -2,70 +2,298 @@
...
+     * @var string|null
+     */
+    private $pass;
+
+    /**
+     * Query string.
+     *
+     * @var string|null
+     */
+    private $query;
```
westonruter, line 94: Eventually it will be nice to add a way to access/manipulate the query vars as an array.

src/Url.php
westonruter: Done in ff5995e and https://github.com/ampproject/amp-wp/pull/6037/commits/178e68c813ffd2a11c49ccc7ad7b4e01b60986b9

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -0,0 +1,375 @@
...
+     * @param string $url Url to check.
+     * @return bool Whether the provided URL uses the AMP Cache.
+     */
+    private function usesAmpCacheUrl($url)
+    {
+        if (! $url) {
+            return false;
+        }
+
+        return strpos($url, Amp::CACHE_HOST) === 0;
```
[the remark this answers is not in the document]
westonruter: Done in e0dd638

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -0,0 +1,356 @@
...
+        $ampUrlPrefix = rtrim($ampUrlPrefix, '/');
+
+        if ($ampRuntimeVersion && $rtv) {
+            $ampUrlPrefix = RuntimeVersion::appendRuntimeVersion($ampUrlPrefix, $ampRuntimeVersion);
+        } elseif ($lts) {
+            $ampUrlPrefix .= '/lts';
+        }
+
+        return $ampUrlPrefix;
+    }
```
westonruter, line 339: I just realized that the `rtv` and `ampRuntimeVersion` are both empty. Instead of seeing the AMP runtime loaded with https://cdn.ampproject.org/rtv/012103261048002/v0.mjs it comes out rtv-less: https://cdn.ampproject.org/v0.mjs

However, if I hack this to force `$rtv` to be `true` and `$ampRuntimeVersion` to be `012103261048002`, then the result is invalid AMP: 

> The attribute 'src' in tag 'amphtml module engine script' is set to the invalid value 'https://cdn.ampproject.org/rtv/012103261048002/v0.mjs'

I guess versioned URLs are not yet valid AMP?

src/Url.php
schlessera: I'd recommend against that. This is supposed to behave like a value object, so it should be immutable to avoid bugs.

It could have a method to retrieve a new clone with a different host, though:
```
$cachedUrl = $url->withHost($cachedHost);
```
The important part is that the original object is not changed (as that would changed it in all the places it was passed around), but rather we return a new instance with the adapted host.

src/Optimizer/Transformer/RewriteAmpUrls.php
schlessera: That would fail the first check (`!empty($url->scheme)`) and generate an error. Looks correct to me.

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -176,7 +176,7 @@ private function usesAmpCacheUrl($url)
             return false;
         }
 
-        return strpos($url, Amp::CACHE_HOST) === 0;
+        return substr_compare($url, Amp::CACHE_HOST, 0, strlen(Amp::CACHE_HOST)) === 0;
```
schlessera, line 179: I don't think this is an improvement. It is both more complex and slower, as I'd guess we never have 10MB URLs to begin with.

src/Url.php
schlessera: That would be powered by a separate object, so that you'd be able to use it as part of a URL, or stand-alone when dealing with WP query vars, for example.

src/Optimizer/Transformer/RewriteAmpUrls.php
westonruter: Only `lts` is valid. RTV-specific scripts are not yet valid: https://github.com/ampproject/amphtml/blob/52407d1c72c5e2051fda427270b9503ae9900ebf/validator/validator-main.protoascii#L3109-L3256

I think this is being tracked by https://github.com/ampproject/amphtml/issues/27546, although this is about self-hosting and not allowing versioned script URLs. @sebastianbenz is that right?

sebastianbenz: Correct, there are no plans to support RTV versions served from cdn.ampproject.org.

src/Optimizer/Transformer/RewriteAmpUrls.php
schlessera: @westonruter Are you okay with reverting this change?

westonruter: Sure, go ahead.

src/Optimizer/Transformer/RewriteAmpUrls.php
westonruter: I mean, if such URLs should be supported.

westonruter (approved):

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -0,0 +1,356 @@
...
+            $preloadNode->setAttribute(Attribute::AS_, Tag::SCRIPT);
+            $preloadNode->setAttribute(Attribute::CROSSORIGIN, Attribute::CROSSORIGIN_ANONYMOUS);
+            $preloadNode->setAttribute(Attribute::HREF, $esmScriptUrl);
+            $preloadNode->setAttribute(Attribute::REL, Attribute::REL_MODULEPRELOAD);
+        }
+
+        $nomoduleNode = $document->createElement(Tag::SCRIPT);
+        $nomoduleNode->addBooleanAttribute(Attribute::ASYNC);
+        $nomoduleNode->addBooleanAttribute(Attribute::NOMODULE);
+        $nomoduleNode->setAttribute(Attribute::SRC, $scriptUrl);
```
westonruter, line 218: Shouldn't the `script[nomodule]` get `crossorigin=anonymous` too? 

```suggestion
        $nomoduleNode->setAttribute(Attribute::SRC, $scriptUrl);
        $nomoduleNode->setAttribute(Attribute::CROSSORIGIN, Attribute::CROSSORIGIN_ANONYMOUS);
```

westonruter: Depends on https://github.com/ampproject/amp-toolbox/pull/1194

schlessera: Merging for now (to get tests to pass in other PRs, and creating a separate issue for this.

schlessera: => https://github.com/ampproject/amp-toolbox-php/pull/140

src/Optimizer/Transformer/RewriteAmpUrls.php
sebastianbenz: Weston is right, we want an explicit https URL.

PR #128, opened 2021-04-06
Calling curl_errno() before closing the connection throws an error

jeffochoa:
An example here:

```text
Amp render failure: Cannot inline the amp-runtime CSS in unspecified version into <style amp-runtime="">: ErrorException: curl_errno(): supplied resource is not a valid cURL handle resource in /vendor/ampproject/amp-toolbox/src/RemoteRequest/CurlRemoteGetRequest.php:136
```

See for reference: [https://stackoverflow.com/questions/19442999/getting-curl-error-2-is-not-a-valid-curl-handle-resource-while-fetching-all-u](https://stackoverflow.com/questions/19442999/getting-curl-error-2-is-not-a-valid-curl-handle-resource-while-fetching-all-u)


src/RemoteRequest/CurlRemoteGetRequest.php
```diff
@@ -146,6 +144,8 @@ static function ($curl, $header) use (&$headers) {
                 continue;
             }
 
+            curl_close($curlHandle);
```
westonruter, line 147: In the case of no body being returned, the `continue` would mean the closing wouldn't happen, right?

How about rather just putting `$curlErrno = curl_errno($curlHandle);` right before `curl_close()`?

westonruter (approved):

schlessera (approved):

schlessera:
Thanks for the PR, @jeffochoa !

PR #130, opened 2021-04-06
Fix deduplication of AMP scripts and improve ReorderHead transformer

### LINKED ISSUE #125: ReorderHead transformer accidentally deduplicates based on URL
The `ReorderHead` transformer uses the URL as an array key when collecting `<script>` elements to reorder. This means that if multiple `<script>` elements use the same URL (like for `.mjs` script + `nomodule` fallback), they overwrite each other in the array and are effectively "deduplicated".

The `ReorderHead` transformer needs a more robust algorithm. It seems that deduplication is still desirable, but will need to rely on "URL + a selection of attributes". The exact selection needs to be clarified, but for now it might help to hash the URL + the presence of `nomodule` as a first bug fix.

westonruter:
Ports (and depends on) https://github.com/ampproject/amp-toolbox/pull/1192. See https://github.com/ampproject/amp-toolbox/issues/1191. Fixes #125.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -12,6 +12,7 @@
 
 	<!-- Exclude Composer vendor folder -->
 	<exclude-pattern>amp-toolbox/vendor/</exclude-pattern>
+	<exclude-pattern>amp-toolbox-php/vendor/</exclude-pattern>
 
 	<!-- Include sniffs for PHP cross-version compatibility. -->
 	<config name="testVersion" value="5.6-"/>
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -11,8 +11,7 @@
 	<exclude-pattern>resources/*</exclude-pattern>
 
 	<!-- Exclude Composer vendor folder -->
-	<exclude-pattern>amp-toolbox/vendor/</exclude-pattern>
-	<exclude-pattern>amp-toolbox-php/vendor/</exclude-pattern>
+	<exclude-pattern>vendor/*</exclude-pattern>
 
 	<!-- Include sniffs for PHP cross-version compatibility. -->
 	<config name="testVersion" value="5.6-"/>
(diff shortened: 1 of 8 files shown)


.phpcs.xml.dist
```diff
@@ -11,7 +11,8 @@
 	<exclude-pattern>resources/*</exclude-pattern>
 
 	<!-- Exclude Composer vendor folder -->
-	<exclude-pattern>vendor/*</exclude-pattern>
+	<exclude-pattern>amp-toolbox/vendor/</exclude-pattern>
+	<exclude-pattern>amp-toolbox-php/vendor/</exclude-pattern>
```
schlessera, line 15: This was probably an unintended change.

src/Amp.php
```diff
@@ -151,7 +151,9 @@ public static function isRuntimeScript(DOMNode $node)
 
         if (
             substr($src, -6) !== '/v0.js'
+            && substr($src, -7) !== '/v0.mjs'
             && substr($src, -14) !== '/amp4ads-v0.js'
+            && substr($src, -15) !== '/amp4ads-v0.mjs'
```
schlessera, line 156: The more comparisons we have here, the more likely it is that a single regex might be faster. Maybe we can just add a TODO here to check performance later on?
```suggestion
			// @TODO Compare performance against single regex.
            substr($src, -6) !== '/v0.js'
            && substr($src, -7) !== '/v0.mjs'
            && substr($src, -14) !== '/amp4ads-v0.js'
            && substr($src, -15) !== '/amp4ads-v0.mjs'
```

schlessera (changes requested):

schlessera:
I'm merging this into the PR #119 to simplify testing in the WP plugin.

.phpcs.xml.dist
westonruter: It was intentional, actual. I found that when I ran phpcs it would math no files when I was working with the package inside of the AMP plugin's vendor directory. The first line prevents this from happening. The second line ensures the expected exclusions work on GHA when the repo is cloned standalone.

schlessera: I already merged without it, and I also never saw this problem before. Can you open an issue for me to look into it then?

westonruter: See #132

PR #136, opened 2021-04-08
Check hero image media attribute before srcset

schlessera:
This PR moves the media attribute check further up to avoid throwing an error for a hero image that was not meant to receive a preload anyway.


westonruter:
Comment is removed as expected:

Before | After
-------|------
image | image

westonruter (approved):

PR #141, opened 2021-04-09
Prevent SSR for amp-img occurring after second paragraph

schlessera:
This PR skips hero image prerendering after paragraph 2 and adds lazy-loading to hero images that were detected without being explicitly marked.

See https://github.com/ampproject/amp-toolbox/pull/1196

Depends on https://github.com/ampproject/amp-toolbox/pull/1198


src/Optimizer/Transformer/PreloadHeroImage.php
```diff
@@ -717,4 +733,32 @@ private function supportsSrcset()
...
+    {
+        while($element) {
+            if (!$element instanceof Element) {
+                $element = $element->parentNode;
+                continue;
+            }
+
+            if ($element->hasAttribute(Attribute::DATA_HERO)) {
+                return true;
+            }
```
westonruter, line 753: Since `DetermineHeroImages` in the AMP plugin is making educated decisions about what images are heroes in WordPress, I think such images should _not_ get `loading=lazy`.

```suggestion
            if ($element->hasAttribute(Attribute::DATA_HERO) || $element->hasAttribute(Attribute::DATA_HERO_CANDIDATE)) {
                return true;
            }
```

src/Optimizer/Transformer/PreloadHeroImage.php
```diff
@@ -563,6 +568,18 @@ private function generateImg(HeroImage $heroImage, Document $document)
         $imgElement->setAttribute(Attribute::CLASS_, self::SSR_IMAGE_CLASS);
         $imgElement->setAttribute(Attribute::DECODING, 'async');
 
+
+        // If the image was detected as hero image candidate (and thus lacks an explicit data-hero), mark it as a hero
+        // and add loading=lazy to guard against making the page performance even worse by eagerly loading an image
+        // outside the viewport.
+        if (! $this->isMarkedAsHeroImage($element)) {
+            $imgElement->setAttribute(Attribute::LOADING, 'lazy');
+        }
```
westonruter, line 577: Maybe there should be an `else` case here to see if there is a `noscript > img` maybe this should copy the `loading` attribute from that image if it is set.

src/Optimizer/Transformer/PreloadHeroImage.php
westonruter: Or maybe not per https://github.com/ampproject/amp-wp/pull/6062/files#r611001630.  Alternatively: https://github.com/ampproject/amp-toolbox-php/pull/141/files#r611001917.

src/Optimizer/Transformer/PreloadHeroImage.php
schlessera: I'm creating a separate issue for this, as I want the tests to be fixed again.
See https://github.com/ampproject/amp-toolbox-php/issues/147

PR #145, opened 2021-04-10
Add autoloader scenario for package as dependency

schlessera:
The autoloader is not properly detected by the binary `amp` right now if the package was included as a dependency, where it is located at `$project/vendor/ampproject/amp-toolbox/bin/amp`.

This PR adds the above scenario.


schlessera:
Test failures are unrelated.

schlessera:
Test failures are unrelated.

PR #146, opened 2021-04-10
Add more dynamic logging behavior

schlessera:
The way logging was handled made for all options parsing errors to be dumped as full traces.

This PR changes the logging to be more dynamic, so that traces are only shown when the `--loglevel` is set to `debug`.


schlessera:
Test failures are unrelated.

PR #150, opened 2021-04-11
Make Configuration extensible

schlessera:
To make https://github.com/ampproject/amp-wp/issues/6064 possible, the `Configuration` object injected into the `TransformationEngine` class needs to be extensible in some form.

This PR turns the current final `Configuration` class into an interface and provides an implementation `DefaultConfiguration` to use as a starting point.

### DIFF
--- a/src/Optimizer/DefaultConfiguration.php
+++ b/src/Optimizer/DefaultConfiguration.php
@@ -54,7 +54,7 @@ class DefaultConfiguration implements Configuration
     public function __construct($configurationData = [])
     {
         $this->configuration = array_merge(
-            Configuration::DEFAULTS,
+            static::DEFAULTS,
             $this->validateConfigurationKeys($configurationData)
         );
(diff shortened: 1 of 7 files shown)


src/Optimizer/DefaultConfiguration.php
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Instantiate a Configuration object.
+     *
+     * @param array $configurationData Optional. Associative array of configuration data to use. This will be merged
+     *                                 with the default configuration and take precedence.
+     */
+    public function __construct($configurationData = [])
+    {
+        $this->configuration = array_merge(
+            Configuration::DEFAULTS,
```
westonruter, line 57: Would this be better so it can be overridden by child classes?

```suggestion
            static::DEFAULTS,
```

westonruter (approved):

PR #167, opened 2021-04-24
Fix covers annotation for CLI exceptions and improve overall coverage

schlessera:


src/Url.php
```diff
@@ -143,7 +143,8 @@ public function __construct($url = null, Url $baseUrl = null)
 
         if ($baseUrl) {
             if (isset($parsedUrl[self::PATH]) && 0 !== strpos($parsedUrl[self::PATH], '/')) {
-                $parsedUrl[self::PATH] = $this->unrelativizePath("{$baseUrl->path}/{$parsedUrl[self::PATH]}");
+                $root                  = rtrim($baseUrl->path, '/');
```
westonruter, line 146: No `untralingsla💩()`?

😉

Issue #17, opened 2020-08-17
Remove preload for AMP runtime again when boilerplate was removed


schlessera:
## Feature description

When the AMP boilerplate is removed by the SSR logic, preloading of the AMP runtime is less critical and it seems to slow down other assets that are of higher priority in that scenario.

The Optimizer should remove an existing preload for the AMP runtime again when the boilerplate was indeed removed.

This was originally reported by @mehigh for an [XWP](https://xwp.co/) project.

/cc @sebastianbenz

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation brief

* Add small transformer that runs after the `ServerSideRendering` transformer and removes an existing AMP runtime preload iff the AMP boilerplate was removed.

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
Given that the test results all seem to indicate this to be a net positive change, I'd like to implement this for the PHP Optimizer. @sebastianbenz, @kristoferbaxter should I also open corresponding issues for the Go and JS versions of the Optimizer?

kristoferbaxter:
We can run a more thorough experiment before moving forward.

Will sync with team today on this.

sebastianbenz:
Are there any updates on this?

jridgewell:
Maybe @kevinkimball or @erwinmombay can run an AMP Cache experiment for this? Should be simple enough.

schlessera:
@sebastianbenz @kristoferbaxter @jridgewell Any updates on this? It's looking like a promising change, so I'd love to implement this for the library and WordPress integration.

westonruter:
I've prototype this along with postponing the loading of scripts that aren't render-critical at https://github.com/ampproject/amp-wp/issues/6182#issuecomment-865363567.

kristoferbaxter:
There is no update beyond the need for statistical analysis at a large enough scale to validate results.

If we choose to adopt this and other suggestions there are several changes required for systems that ingest AMP documents and validate them.

kristoferbaxter:
We should move forward with removal for now, data suggests the origin performance can be better without, but the preload version from a cache will continue to insert and perform better with in that scenario.

robin-scott:
[...]

PR #179, opened 2021-05-03
Add resource hints transformers

schlessera:
This PR adds two transformers to the Optimizer:

1. `AmpRuntimePreloads`

	This adds preloads for the AMP runtime scripts (WIP) and styles, if they are needed.
 
2. `GoogleFontsPreconnect`

	This adds a preconnect for the Google Fonts domain if Google Fonts are being used, and adds a DNS prefetch as a fallback.

It also adds an abstraction for managing resource hints via the `Dom\Document`:
image

TODOS:

- [ ] When should the AMP runtime script be preloaded (if at all)?
- [ ] Should it be possible to use `preconnect` without the best-practice-advised `dns-prefetch`?
- [ ] Should we add other resource hints right away, even if we don't need them, i.e. `prefetch`, `prerender`, 'dns-prefetch` on its own...?


src/Dom/ResourceHintManager.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Add a link tag.
+     *
+     * @param string   $rel        A 'rel' string.
+     * @param string   $href       URL to link to.
+     * @param string[] $attributes Associative array of attributes and their values.
+     */
+    public function addLinkTag($rel, $href, $attributes = [])
```
westonruter, line 88: This should perhaps be `private` as otherwise it a user could add arbitrary link relations via this manager, including `stylesheet`, `glossary`, etc.

This being the case, should there be additional method for `modulepreload`? Should there be one for `dns-prefetch`? There seems to be value in adding `dns-prefetch` _without_ also adding `preconnect` for non-critical resources per [MDN](https://developer.mozilla.org/en-US/docs/Web/Performance/dns-prefetch#:~:text=if%20a%20page%20needs%20to%20make%20connections%20to%20many%20third-party%20domains):

<blockquote>
<p><strong>Note</strong>: If a page needs to make connections to many third-party domains, preconnecting them all is counterproductive. The <code>preconnect</code> hint is best used for only the most critical connections. For the others, just use <code>&lt;link rel="dns-prefetch"&gt;</code> to save time on the first step — the DNS lookup.</p>
</blockquote>

westonruter: Additional methods would be to add `prefetch` and `prerender` links.

src/Dom/ResourceHintManager.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Add a preconnect resource hint.
+     *
+     * @param string      $href  URL to link to.
+     * @param string      $type  Optional. Type of the resource. Defaults to type 'image'.
+     * @param string|null $media Optional. Media query to add to the preload. Defaults to none.
+     */
+    public function addPreload($href, $type = RequestDestination::IMAGE, $media = null)
```
westonruter, line 69: Technically preload links are not a [resource hints](https://www.w3.org/TR/resource-hints/). It's a [separate thing](https://www.w3.org/TR/preload/). See https://core.trac.wordpress.org/ticket/42438#comment:3

This this reason the note about `private` below, what about changing `ResourceHintManager` to be rather `LinkManager`. Then this could be used for adding any `link`.

src/Optimizer/Transformer/AmpRuntimePreloads.php
```diff
@@ -0,0 +1,85 @@
...
+     * Check whether the AMP runtime script is needed.
+     *
+     * @param Document $document Document to check in.
+     * @return bool Whether the AMP runtime script is needed.
+     */
+    private function isAmpRuntimeScriptNeeded(Document $document)
+    {
+        // TODO: What condition should be used here... same as isAmpRuntimeCssNeeded()?
+        return false;
+    }
```
westonruter, line 46: This is locked by https://github.com/ampproject/amp-toolbox-php/issues/17, yes?

src/Optimizer/Transformer/GoogleFontsPreconnect.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+final class GoogleFontsPreconnect implements Transformer
+{
+
+    /**
+     * Domain that the Google Fonts static files are loaded from.
+     *
+     * @var string
+     */
+    const GOOGLE_FONTS_STATIC_DOMAIN = 'https://fonts.gstatic.com';
```
westonruter, line 22: Since to me `domain` would be just the host like `fonts.gstatic.com`

```suggestion
    const GOOGLE_FONTS_STATIC_ORIGIN = 'https://fonts.gstatic.com';
```

src/Optimizer/Transformer/GoogleFontsPreconnect.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Domain that the Google Fonts API is accepting requests from.
+     *
+     * @var string
+     */
+    const GOOGLE_FONTS_API_DOMAIN = 'https://fonts.googleapis.com/';
```
westonruter, line 29: Doesn't work as well as above, so perhaps `BASE_URL`?

```suggestion
    const GOOGLE_FONTS_API_BASE_URL = 'https://fonts.googleapis.com/';
```

westonruter (changes requested):

src/Optimizer/Transformer/GoogleFontsPreconnect.php
```diff
@@ -0,0 +1,67 @@
+<?php
```
pierlon, line 1: Rather than providing a transformer for one custom font, would we want to create a generic `FontsPreconnect` transformer that supports adding resource hints for any of the custom fonts on the allowlist?

(See related [validation rule](https://github.com/ampproject/amphtml/blob/f2a49b8efa4ecb0cd1dd273977799299a9e1f816/validator/validator-main.protoascii#L349-L391))

westonruter: This may not be necessary because I think the unique issue with Google fonts is that the font files are located on `https://fonts.gstatic.com` whereas the stylesheets that reference the fonts are on `https://fonts.googleapis.com`, so adding the preconnect to `https://fonts.gstatic.com` will ensure that the font files are loaded fast once the stylesheet is loaded. 

In contrast, looking at Font Awesome, I see the font stylesheet example of https://use.fontawesome.com/releases/v5.8.1/css/all.css references font files which are on the same domain. Therefore no preconnect is needed in that case.

src/Dom/ResourceHintManager.php
```diff
@@ -0,0 +1,107 @@
...
+    public function __construct(Document $document)
+    {
+        $this->document = $document;
+    }
+
+    /**
+     * Add a preconnect resource hint.
+     *
+     * @param string $href        URL to link to.
+     * @param bool   $crossorigin Whether the link is crossorigin.
```
westonruter, line 47: The `$crossorigin` isn't actually a bool, is it? It can have the values `anonymous` or `use-credentials`. If empty then it is the same as `anonymous`: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin

src/Optimizer/Transformer/AmpRuntimePreloads.php
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param Document        $document DOM document to apply the transformations to.
+     * @param ErrorCollection $errors   Collection of errors that are collected during transformation.
+     * @return void
+     */
+    public function transform(Document $document, ErrorCollection $errors)
+    {
+        if ($this->isAmpRuntimeScriptNeeded($document)) {
+            $document->links->addPreload($this->getAmpRuntimeScriptHost(), RequestDestination::SCRIPT);
+        }
```
westonruter, line 29: This is currently being handled by the `RewriteAmpUrls` transformer. Should the logic be moved here? At the very least it seems `RewriteAmpUrls` should make use of the `LinkManager` in its `createPreload` method.

schlessera: I made `RewriteAmpUrls` reuse the link manager. I also extended the link manager so that it ensure no duplicate links are being produced. This currently uses both `href` and `rel` for the uniqueness key.

westonruter (approved):
LGTM, other than a couple minor things that can be addressed now or later.

src/Dom/LinkManager.php
```diff
@@ -0,0 +1,314 @@
...
+
+        if ($this->referenceNode) {
+            $link = $this->document->head->insertBefore($link, $this->referenceNode->nextSibling);
+        } else {
+            $link = $this->document->head->appendChild($link);
+        }
+
+        if (! $link instanceof Element) {
+            throw FailedToCreateLink::forLink($link);
+        }
```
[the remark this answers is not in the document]
schlessera: If you do something fishy with the document, you might end up with a `DOMElement` that is not an `Element`.

Also, PhpStan is happier with this restriction.

src/Optimizer/Transformer/RewriteAmpUrls.php
```diff
@@ -228,31 +214,23 @@ private function addEsm(Document $document, Element $scriptNode)
...
 
-        $preloadNode = $document->createElement(Tag::LINK);
-
-        $preloadNode->setAttribute(Attribute::REL, Attribute::REL_PRELOAD);
-        $preloadNode->setAttribute(Attribute::HREF, $href);
-        $preloadNode->setAttribute(Attribute::AS_, $type);
-
-        return $preloadNode;
+        // TODO: Should the preloads be crossorigin here? Maybe conditionally based on host vs origin?
+        $document->links->addPreload($href, $type, null, false);
```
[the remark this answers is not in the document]
schlessera: Created separate issue for this: https://github.com/ampproject/amp-toolbox-php/issues/233

src/Dom/LinkManager.php
```diff
@@ -0,0 +1,314 @@
...
+                $node->tagName !== Tag::LINK
+            ) {
+                $node = $nextSibling;
+                continue;
+            }
+
+            $href = $node->getAttribute(Attribute::HREF);
+            $rel  = $node->getAttribute(Attribute::REL);
+
+            $key = $this->getKey($node);
```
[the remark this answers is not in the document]
schlessera: Resolved via https://github.com/ampproject/amp-toolbox-php/pull/179/commits/c3a2159200019fb9df2ee5df3431569660e0fb15

westonruter (approved):

PR #183, opened 2021-05-05
Skip detecting ASCII when charset is set to "US-ASCII"

06romix:
Problem: ASCII has an alternative name - US-ASCII but, PHP function mb_list_encodings knows only ASCII and sanitizes this encoding.

Solution: add US-ASCII to \AmpProject\Dom\Document\Encoding::MAPPINGS

### DIFF
--- a/src/Dom/Document/Encoding.php
+++ b/src/Dom/Document/Encoding.php
@@ -39,7 +39,7 @@ interface Encoding
         // Assume ISO-8859-1 for some charsets.
         'latin-1' => 'ISO-8859-1',
         // US-ASCII is one of the most popular ASCII names and used as HTML charset,
-        // but mb_list_encodings contains only "ASCII"
+        // but mb_list_encodings contains only "ASCII".
         'us-ascii' => 'ascii',
     ];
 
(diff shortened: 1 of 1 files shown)


src/Dom/Document/Encoding.php
```diff
@@ -38,6 +38,9 @@ interface Encoding
     const MAPPINGS = [
         // Assume ISO-8859-1 for some charsets.
         'latin-1' => 'ISO-8859-1',
+        // US-ASCII is one of the most popular ASCII names and used as HTML charset,
+        // but mb_list_encodings contains only "ASCII"
```
schlessera, line 42: This library originated from within a WordPress plugin, so we have some consistency rules that come from the WordPress Coding Standards (WPCS). One of these rules is that all comments are terminated with a period. We're keeping this rule around for consistency.
(And I now notice that we are not actually checking these rules atm...)
```suggestion
        // but mb_list_encodings contains only "ASCII".
```

schlessera (changes requested):

src/Dom/Document/Encoding.php
schlessera: I created a new issue for this, as ideally all such rules should be enforced via code quality checks: #184

06romix: Corrected the comment.

schlessera (approved):

schlessera:
Thanks for the PR, @06romix !

PR #191, opened 2021-05-11
Remove PHPCS and replace with ECS

schlessera:
This PR removes PHPCS and replaces it with ECS - Easy Coding Standard.

This has the following advantages:
- Use of both PHPCS and PHPCSF via one PHP config file
- Advanced caching for quick rechecking during development
- Possibility to check code snippets in markdown files


schlessera:
Something to note: the original idea of trying to move to ECS is that it will allow the code generation for the validator spec to be run through the ECS fixed, which reliably enforces the exact same code style than for the rest of the code. This is unfortunately not easily done with PHPCS, as its fixing only covers a part of its ruleset.

westonruter (approved):
LGTM, though I've never used ECS before

Issue #24, opened 2020-11-12
Detect malformed byte sequences and warn the user


schlessera:
The `Dom\Document` abstraction relies on valid byte sequences to either keep existing UTF-8 content or convert content in a different encoding to the UTF-8 encoding.

If a document contains a malformed byte sequence, the parsing done by `DOMDocument` fails in a random way (silently) and produces garbage:
![image](https://user-images.githubusercontent.com/83631/98935201-768c1480-24e3-11eb-9765-c174ae3ff3dd.png)

This is not only bad because of the garbled end result, it also looks like the AMP conversion is the culprit and makes it difficult for the user to diagnose the real issue.

The source of this problem (the malformed byte sequence) can actually be detected in the non-AMP document and is also shown by the W3C validator:
![Showing results for __news androidlist jp_2020_11_12__E3_81_99_E3_81_  2020-11-12 at 11 58 47 AM](https://user-images.githubusercontent.com/83631/98935291-9a4f5a80-24e3-11eb-96f3-45360b7d1eee.png)

PHP has a function `mb_check_encoding()` which seems to be able to detect malformed byte sequences (see https://3v4l.org/ciAPf). We should investigate whether we can detect this problem upfront and warn the user about it.

Things to check:
- How reliable does `mb_check_encoding()` work, and does it get trigger on other issues as well?
- What is the performance impact of checking the entire document before parsing it via `Dom\Document`?

westonruter:
Was the issue caused by truncating text via `substr()` instead of `mb_substr()`?

schlessera:
I assume so. The user didn't get back to me on what the actual code was that broke the encoding, but `substr` is the most likely culprit.

06romix:
> How reliable does mb_check_encoding() work, and does it get trigger on other issues as well?

On PHP documentation there is [the user note](https://www.php.net/manual/en/function.mb-check-encoding.php#89286) with test detecting bad byte stream and bad byte sequence. But now I mb_check_encoding works correctly on different PHP versions. So I think `mb_check_encoding` can detect both bad byte stream and bad byte sequence.

> What is the performance impact of checking the entire document before parsing it via Dom\Document?

I added `mb_check_encoding` to `\AmpProject\Dom\Document::loadHTML` and parse html took 10% more time.
Tested 10 times with 10000 parsings using `tests/spec/valid-amp/files/amp-selector.html`

schlessera:
[...]

PR #200, opened 2021-05-15
Add option to detect malformed byte sequences

06romix:
This PR adds the option for detecting malformed byte sequences according to #24

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -8,6 +8,7 @@ use AmpProject\DevMode;
 use AmpProject\Dom\Document\Encoding;
 use AmpProject\Dom\Document\Option;
 use AmpProject\Exception\FailedToRetrieveRequiredDomElement;
+use AmpProject\Exception\InvalidByteSequence;
 use AmpProject\Exception\MaxCssByteCountExceeded;
 use AmpProject\Optimizer\CssRule;
 use AmpProject\Tag;
@@ -2154,15 +2155,15 @@ final class Document extends DOMDocument
      * If invalid byte sequences are passed to `DOMDocument`, it fails silently and produces Mojibake.
      *
      * @param string $source The HTML fragment string.
-     * @throws \InvalidArgumentException If $source contains invalid byte sequences.
+     * @throws InvalidByteSequence If $source contains invalid byte sequences.
      */
     private function detectInvalidByteSequences($source)
     {
-        if ($this->options[Option::CHECK_ENCODING] && function_exists('mb_check_encoding') && ! mb_check_encoding($source)) {
-            throw new \InvalidArgumentException(
-                'Provided HTML contains invalid byte sequences. ' .
-                'This is usually fixed by replacing string manipulation functions with their `mb_*` multibyte counterparts.'
-            );
+        if ($this->options[Option::CHECK_ENCODING]
+            && function_exists('mb_check_encoding')
+            && ! mb_check_encoding($source)
+        ) {
+            throw InvalidByteSequence::forHtml();
         }
     }
 }
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -2159,7 +2159,8 @@ final class Document extends DOMDocument
      */
     private function detectInvalidByteSequences($source)
     {
-        if ($this->options[Option::CHECK_ENCODING]
+        if (
+            $this->options[Option::CHECK_ENCODING]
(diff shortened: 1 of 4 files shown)


src/Dom/Document.php
```diff
@@ -2145,4 +2147,22 @@ private function addHttpEquivCharset($html)
...
+     *
+     * @param string $source The HTML fragment string.
+     * @throws \InvalidArgumentException if source contains some malformed byte sequences
+     */
+    private function assertNoneMalformedByteSequences($source)
+    {
+        if ($this->options[Option::ADDITIONAL_VALIDATION] && ! mb_check_encoding($source)) {
+            throw new \InvalidArgumentException(
+                'Passed html contains some malformed byte sequences. ' .
+                'Perhaps you use function "substr" instead of "mb_substr".'
```
schlessera, line 2164:
```suggestion
                'Provided HTML contains invalid byte sequences. ' .
                'This is usually fixed by replacing string manipulation functions with their `mb_*` multibyte counterparts.'
```

src/Dom/Document.php
```diff
@@ -2145,4 +2147,22 @@ private function addHttpEquivCharset($html)
...
+     *
+     * If passed bad byte sequences DOMDocument would fail in a random way (silently) and produces garbage.
+     *
+     * @param string $source The HTML fragment string.
+     * @throws \InvalidArgumentException if source contains some malformed byte sequences
+     */
+    private function assertNoneMalformedByteSequences($source)
+    {
+        if ($this->options[Option::ADDITIONAL_VALIDATION] && ! mb_check_encoding($source)) {
+            throw new \InvalidArgumentException(
```
schlessera, line 2162: This should use a custom exception `InvalidByteSequence` with a named constructor (see `src/Exception`).

schlessera (changes requested):
I think the `mb_check_encoding()` will detect any invalid byte sequence, so the language used should move from "malformed" (which is a specific issue) to "invalid" (which covers all detected issues).

I've provided change suggestions for the code, but the tests will also needed to be adapted to match the adapted code.

src/Dom/Document.php
06romix: Should the exception message contain an HTML source?
If not, is there necessary to write a test for InvalidByteSequence that contains only a message?

06romix:
I'm with you on that, "invalid" is a good choice for `mb_check_encoding()`.

Thank you for your detailed review. I appreciate it.

src/Dom/Document.php
schlessera: Normally yes, but for string-based HTML, it is actually hard to produce something useful for an exception message that doesn't flood the screen/logs... Not sure how we would go about that here.

06romix: The third way is to add a second argument to the named constructor and then provide a method like `$invalidByteSequence->getHtmlSource()` to get the html contained invalid byte sequence.

tests/Dom/DocumentTest.php
```diff
@@ -1280,4 +1281,34 @@ public function testCreateElementWithAttributes()
             $document->saveHTMLFragment($element)
         );
     }
+
+    /**
+     * Test asserting none malformed byte sequences
+     *
+     * @covers \AmpProject\Dom\Document::detectInvalidByteSequences()
```
schlessera, line 1288:
```suggestion
     * @covers \AmpProject\Dom\Document::detectInvalidByteSequences()
     * @covers \AmpProject\Exception\InvalidByteSequence::forHtml()
```

tests/Dom/DocumentTest.php
```diff
@@ -1280,4 +1281,34 @@ public function testCreateElementWithAttributes()
...
+        $document::fromHtmlFragment(
+            "<div>Some text \x00\x81<div>",
+            [Option::CHECK_ENCODING => true]
+        );
+    }
+
+    /**
+     * Test asserting none bad byte sequences
+     *
+     * @covers \AmpProject\Dom\Document::detectInvalidByteSequences()
```
schlessera, line 1303:
```suggestion
     * @covers \AmpProject\Dom\Document::detectInvalidByteSequences()
     * @covers \AmpProject\Exception\InvalidByteSequence::forHtml()
```

schlessera (changes requested):

schlessera:
Thanks for the PR, @06romix !

Issue #107, opened 2021-03-22
Inline SVG attributes get lowercased


dritter:
Hi there,

the optimizer does not seem to respect the XML Namespace. As SVG is XML, and thus case sensitive (opposed to HTML, which is case insensitive), the [SVG viewBox attribute](https://www.w3.org/TR/SVG2/coords.html#ViewBoxAttribute) gets lowercased by the amp-optimizer. Browsers seem to ignore the problem, so this is more a nitpick.

See the `viewBox` attribute here:

Before:
```svg
[...]

PR #210, opened 2021-05-25
Add AutoExtensions transformer

### LINKED ISSUE #32: New Transformer: AutoExtensions
- [ ] Automatically add scripts for used extensions
- [ ] Strip unused extension scripts

schlessera:
This PR adds a new transformer that automatically imports missing extension scripts and removes unneeded ones.

Depends on #100 
Fixes #32


westonruter:
Should this add a special case for `amp-access-scroll`? Ref https://github.com/ampproject/amp-wp/issues/5999.

schlessera:
Waiting on https://github.com/ampproject/amp-toolbox/pull/1241

westonruter:
@schlessera I have a request: please include a configuration option to _omit_ certain scripts from ever being imported. The use case is present in https://github.com/ampproject/amp-wp/pull/6527 where if the user opts-in to native `POST` forms, we need to prevent the `amp-form` script from being added to the page.

Here's how the plugin is currently handling that omission: https://github.com/ampproject/amp-wp/blob/ebb400816a34c3a9cea768486ac2d19c8211e5cf/includes/class-amp-theme-support.php#L1535-L1542

So if the transformer could be configured to omit `amp-form` or else it could otherwise by default omit `amp-form` if there is any `form[action][method=post]` on the page (since the presence of the `amp-form` extension blocks such forms from working entirely).

schlessera:
@westonruter Yes, makes sense. We can add an `ignore` option that takes an array of extension names. The code would then neither add-if-missing nor remove-if-unneeded that extension script.

ediamin:
Hi @schlessera I've pushed the new changes. The spec tests for AutoExtensions are passing now except the `AutoExtensions - ignore-mask-in-svgs` test. This is related to the https://github.com/ampproject/amp-toolbox-php/issues/107 and https://github.com/ampproject/amp-wp/issues/2045.

Also, added the `ignore` configuration option and unit test cases for the `AutoExtensionsConfiguration` class. We can add ignore list like this,
```php
$config = new AutoExtensionsConfiguration([
    'ignore' => ['amp-form', 'amp-ad', ...]
]);
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,584 @@
...
+            $extensionScripts = $this->maybeAddExtension($document, $extensionScripts, Extension::BIND);
+        }
+
+        return $extensionScripts;
+    }
+
+    /**
+     * Get all attributes for a tag.
+     *
+     * @param string $tagName Name of the node tag.
```
schlessera, line 385:
```suggestion
	 * @TODO: Provide this functionality as part of the validator spec implementation.
	 *
     * @param string $tagName Name of the node tag.
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,584 @@
...
+            // Check for amp-bind attribute bindings.
+            if (strpos($attribute->name, '[') === 0 || strpos($attribute->name, Amp::BIND_DATA_ATTR_PREFIX) === 0) {
+                $usesAmpBind = true;
+            }
+
+            /**
+             * EXPERIMENTAL FEATURE: Rewrite short-form `bindtext` to `data-amp-bind-text` to avoid false-positives we
+             * check for each tag only the supported bindable attributes (e.g. for a div only bindtext, but not
+             * bindvalue).
+             */
```
schlessera, line 345: Rephrasing this a bit to read better. Also note that this is not a docblock, as it doesn't document a specific element. Multi-line comments start with `/*`, not with `/**`.
```suggestion
            /*
             * EXPERIMENTAL FEATURE: Rewrite short-form `bindtext` to `data-amp-bind-text`.
             *
             * To avoid false-positives, we only check the supported bindable attributes for each tag (e.g. for a div
             * only bindtext, but not bindvalue).
             */
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,584 @@
...
+    /**
+     * Remove unneeded extension scripts.
+     *
+     * @param Document  $document         Document to scan for unneeded extensions.
+     * @param Element[] $extensionScripts Array of extension scripts to check.
+     * @return Element[] Adapted array of extension scripts.
+     */
+    private function removeUnneededExtensions(Document $document, $extensionScripts)
+    {
+        // TODO: Add logic.
```
schlessera, line 420: This part is still missing the implementation. It should probably also be configurable via a flag, so you can choose whether removal happens or not.

schlessera (commented):
There is still part of the implementation missing => removal of unneeded extensions.

Also, after getting past the SVG case inconsistency and re-enabling the SVG test, it produced a different failure that needs to be looked into.

src/Optimizer/Transformer/AutoExtensions.php
ediamin: Aren't ignoredExtensions and removeExtensions the same? Cause we are checking the ignoredExtensions list before adding that extension into `extensionScripts` list. For example if we want to remove `amp-form` extension, we can put this into ignoredExtensions list. Please share your thoughts on this.

ediamin:
> Also, after getting past the SVG case inconsistency and re-enabling the SVG test, it produced a different failure that needs to be looked into.

@schlessera The test is failing because of self closing `path` and `g` tags. I can add them into `AmpProject\Tag::SELF_CLOSING_TAGS` list, but the problem is I found out that most of the time the `g` tag is not self closing, so not sure how to handle this scenario. Please suggest how should I approach.

src/Optimizer/Transformer/AutoExtensions.php
schlessera: Ignoring an extension means that the system should not check what to do with that extension. If the extension script is present, it stays in (whether it is needed or not). If the extension script is not present, it is not being added (whether it is needed or not).

Removing unneeded extensions on the other hand is about doing cleanup in case their are left-overs. For an existing document that might already have extension scripts, you can have both missing extension scripts (the author forgot one) and superfluous extension scripts (the author added one, but later removed the actual markup but forgot to remove the corresponding extension script as well). So we are already adding the extension scripts that are missing, but we are not yet removing the extension scripts that are unused.

Ignore => do not even do a check
Remove => strip if unused

schlessera:
I don't think the problem is with the spec of the `g` tag.

Looking at the test output, it seems like the trailing slash for the self-closing tags is being stripped already from the input (i.e. the expected markup shows `<path ...>` instead of `<path .../>`). You should verify that these are kept intact throughout the execution, and then I think this test will also pass.

westonruter:
> Removing unneeded extensions on the other hand is about doing cleanup in case their are left-overs. For an existing document that might already have extension scripts, you can have both missing extension scripts (the author forgot one) and superfluous extension scripts (the author added one, but later removed the actual markup but forgot to remove the corresponding extension script as well). So we are already adding the extension scripts that are missing, but we are not yet removing the extension scripts that are unused.
> 
> Ignore => do not even do a check
> Remove => strip if unused

I wanted to call out another scenario to make sure it is accounted for. In the AMP plugin, we use `amp-carousel` v0.2 even though the `latest` version is still at v0.1. If you use `amp-lightbox-gallery` on a page and do not have the `amp-carousel` script on the page, then the `latest` version of `amp-carousel` is loaded: v0.1. This can result in unexpected behavior: https://github.com/ampproject/amp-wp/pull/3115#issuecomment-890130430.

Therefore, the AMP plugin's auto-extension logic will prevent removal of `amp-carousel` script when there is no `<amp-carousel>` on the page, when the `amp-lightbox-gallery` script is on the page. See 
https://github.com/ampproject/amp-wp/pull/6509. This ensures that the same version of `amp-carousel` is used always. For AutoExtensions transformer, I believe this is an example of a hard-coded dependency which isn't captured in the validator. 

In short: Adding `amp-lightbox-gallery` shouldn't automatically add the `amp-carousel` script, but if the `amp-carousel` script happens to be on the page then don't remove it (even when there is no `<amp-carousel>`.

ediamin:
@schlessera Updated the code with following changes,
1. Add a new filter `SelfClosingSVGElements` to handle the SVG related self closing elements.
2. Add logic to remove the unneeded extension with configuration option. Also added `amp-carousel` as protected extension so it'll never remove even when there is no `<amp-carousel>` in the HTML.

src/Optimizer/Error/CannotParseJsonData.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace AmpProject\Optimizer\Error;
+
+use AmpProject\Dom\Element;
+use AmpProject\Dom\ElementDump;
+use AmpProject\Optimizer\Error;
+use Exception;
+
+final class CannotParseJsonData implements Error
```
schlessera, line 10: Missing class docblock.

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+use AmpProject\Optimizer\Transformer;
+use AmpProject\Optimizer\TransformerConfiguration;
+use AmpProject\Tag;
+use AmpProject\Validator\Spec;
+use AmpProject\Validator\Spec\AttributeList\GlobalAttrs;
+use AmpProject\Validator\Spec\SpecRule;
+use Exception;
+use InvalidArgumentException;
+
+final class AutoExtensions implements Transformer
```
schlessera, line 23: Missing class docblock.

src/Optimizer/Configuration/AutoExtensionsConfiguration.php
```diff
@@ -0,0 +1,183 @@
...
+     * Configuration key that specifies the AMP format.
+     *
+     * @var string
+     */
+    const FORMAT = 'format';
+
+    /**
+     * Configuration key that can disable the automatic importing of extension.
+     *
+     * @var bool.
```
schlessera, line 34: The `@var` here defines the type of the constant, not the type of the value that this configuration key refers to.
```suggestion
     * @var string
```

src/Optimizer/Configuration/AutoExtensionsConfiguration.php
```diff
@@ -0,0 +1,183 @@
...
+     */
+    const AUTO_EXTENSION_IMPORT = 'autoExtensionImport';
+
+    /**
+     * Configuration key that enables experimental conversion of bind attributes.
+     *
+     * @var bool
```
schlessera, line 41: The `@var` here defines the type of the constant, not the type of the value that this configuration key refers to.
```suggestion
     * @var string
```

src/Optimizer/Configuration/AutoExtensionsConfiguration.php
```diff
@@ -0,0 +1,183 @@
...
+     */
+    const EXPERIMENT_BIND_ATTRIBUTE = 'experimentBindAttribute';
+
+    /**
+     * Configuration key that allows individual configuration of extension versions.
+     *
+     * @var array
```
schlessera, line 48: The `@var` here defines the type of the constant, not the type of the value that this configuration key refers to.
```suggestion
     * @var string
```

src/Optimizer/Configuration/AutoExtensionsConfiguration.php
```diff
@@ -0,0 +1,183 @@
...
+     */
+    const EXTENSION_VERSIONS = 'extensionVersions';
+
+    /**
+     * An array of extension names that will not auto import.
+     *
+     * @var string[]
```
schlessera, line 55: The `@var` here defines the type of the constant, not the type of the value that this configuration key refers to.
```suggestion
     * @var string
```

src/Optimizer/Configuration/AutoExtensionsConfiguration.php
```diff
@@ -0,0 +1,183 @@
...
+     */
+    const IGNORED_EXTENSIONS = 'ignoredExtensions';
+
+    /**
+     * An array of extension names that will not auto import.
+     *
+     * @var bool
```
schlessera, line 62: The `@var` here defines the type of the constant, not the type of the value that this configuration key refers to.
```suggestion
     * @var string
```

src/Optimizer/Configuration/AutoExtensionsConfiguration.php
```diff
@@ -0,0 +1,183 @@
...
+                        gettype($value)
+                    );
+                }
+
+                // Assert that the extension names in the ignore list are valid extensions.
+                $reflection = new ReflectionClass(Extension::class);
+                $constants = $reflection->getConstants();
+
+                foreach ($value as $extension) {
+                    if (! in_array($extension, $constants)) {
```
schlessera, line 163: Array functions should always be used in strict mode.
```suggestion
                    if (! in_array($extension, $constants, true)) {
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+        if (! $node->hasAttributes()) {
+            return $extensionScripts;
+        }
+
+        // Look for forms.
+        if ($node->tagName === Tag::FORM) {
+            $extensionScripts = $this->maybeAddExtension($document, $extensionScripts, Extension::FORM);
+        }
+
+        // true if we need to import amp-bind.
```
schlessera, line 312:
```suggestion
        // Will be set to true if we need to import amp-bind.
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+     * @param Element[] $extensionScripts Array of preexisting extension scripts.
+     * @return Element[] Adapted array of extension scripts that includes the previously missing ones.
+     */
+    private function addRequiredExtensionByAttributes(Document $document, Element $node, $extensionScripts)
+    {
+        if (! $node->hasAttributes()) {
+            return $extensionScripts;
+        }
+
+        // Look for forms.
```
schlessera, line 307:
```suggestion
        // The amp-form extension extends the regular <form> tag, so we manually look for that tag.
        // See https://amp.dev/documentation/components/amp-form/.
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+        if ($node->tagName === Tag::FORM) {
+            $extensionScripts = $this->maybeAddExtension($document, $extensionScripts, Extension::FORM);
+        }
+
+        // true if we need to import amp-bind.
+        $usesAmpBind = false;
+
+        // Associative array of new attribute name with data-amp-bind- prefix and attribute value pair.
```
schlessera, line 315:
```suggestion
        // Associative array of new attribute names with data-amp-bind- prefix and attribute value pair.
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+
+            // Element attributes that require amp extensions (e.g. amp-video[dock]).
+            if (! empty($nodeAttributeList[$attribute->name][SpecRule::REQUIRES_EXTENSION])) {
+                $requiresExtensions = $nodeAttributeList[$attribute->name][SpecRule::REQUIRES_EXTENSION];
+                foreach ($requiresExtensions as $requiresExtension) {
+                    $extensionScripts = $this->maybeAddExtension($document, $extensionScripts, $requiresExtension);
+                }
+            }
+
+            // Attribute that requires AMP components (e.g. amp-fx).
```
schlessera, line 341:
```suggestion
            // Attributes that require AMP components (e.g. amp-fx).
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+     * Maybe add a required extension to the list of extension scripts.
+     *
+     * @param Document $document          Document to render the extension scripts into.
+     * @param array    $extensionScripts  Existing list of extension scripts.
+     * @param string   $requiredExtension Required extension to check for.
+     * @return array Adapted list of extension scripts.
+     */
+    private function maybeAddExtension(Document $document, $extensionScripts, $requiredExtension)
+    {
+        if (in_array($requiredExtension, $this->configuration->get(AutoExtensionsConfiguration::IGNORED_EXTENSIONS))) {
```
schlessera, line 528: Array functions should always be used in their strict mode:
```suggestion
        if (in_array($requiredExtension, $this->configuration->get(AutoExtensionsConfiguration::IGNORED_EXTENSIONS), true)) {
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+            return $extensionScripts;
+        }
+
+        if (!array_key_exists($requiredExtension, $extensionScripts)) {
```
schlessera, line 532:
```suggestion
        if (! array_key_exists($requiredExtension, $extensionScripts)) {
```

tests/Optimizer/Configuration/AutoExtensionsConfigurationTest.php
```diff
@@ -0,0 +1,166 @@
...
+    {
+        $this->expectError();
+        $this->expectErrorMessage(
+            'Undefined property: AmpProject\Optimizer\Configuration\AutoExtensionsConfiguration::$nonsense'
+        );
+        $configuration = new AutoExtensionsConfiguration([]);
+        $configuration->nonsense;
+    }
+
+    public function testInvalidConfigurationValueForUnknownAMPFormat()
```
schlessera, line 145:
```suggestion
    public function testInvalidConfigurationValueForUnknownAmpFormat()
```

schlessera:
I just send the review as a "comment", instead of a "change request", but most of the above needs changes, though. But it's only minor tweaks, the overall logic LGTM.

schlessera:
Ah, just realized that I was not able to "Request Changes" because I created the initial PR myself. I've requested a review from you instead now.

tests/Optimizer/Transformer/AutoExtensionsTest.php
```diff
@@ -0,0 +1,193 @@
...
+                TestMarkup::DOCTYPE . '<html><head>' . TestMarkup::META_CHARSET
+                . '<script async custom-element="amp-anim" src="https://cdn.ampproject.org/v0/amp-anim-0.1.js"></script>'
+                . '</head><body><amp-anim></amp-anim></body></html>',
+                [],
+                [
+                    AutoExtensionsConfiguration::REMOVE_UNNEEDED_EXTENSIONS => true,
+                ]
+            ],
+
+            'keep unneeded amp-carousel extension when removeUnneededExtensions is true' => [
```
schlessera, line 154: `amp-carousel` should only be kept if `amp-lightbox-gallery` is present (even if no `amp-carousel` tag is found).

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+    /**
+     * List of extensions added by the maybeAddExtension method.
+     *
+     * @var array
+     */
+    private $addedExtensions = [];
+
+    /**
+     * List of extensions that should not be removed even when there is no usage in the HTML.
+     *
```
schlessera, line 65:
```suggestion
     * List of extensions that should not be removed even when there is no usage in the HTML.
     *
     * The array key is the extension to protect, the array value is an array of extensions that
     * make up the requirements for protecting the extension. An empty array means the extension
     * is protected unconditionally.
     *
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+     * @var array
+     */
+    private $protectedExtensions = [
+        'amp-carousel',
```
schlessera, line 69:
```suggestion
        'amp-carousel' => ['amp-lightbox-gallery'],
```

src/Optimizer/Transformer/AutoExtensions.php
```diff
@@ -0,0 +1,609 @@
...
+    private function removeUnneededExtensions($extensionScripts)
+    {
+        if (! $this->configuration->get(AutoExtensionsConfiguration::REMOVE_UNNEEDED_EXTENSIONS)) {
+            return $extensionScripts;
+        }
+
+        return array_filter($extensionScripts, function ($extension) {
+            return array_key_exists($extension, $this->addedExtensions)
+                || in_array($extension, $this->protectedExtensions);
+        }, ARRAY_FILTER_USE_KEY);
```
schlessera, line 445: This logic needs to be adapted to only protect extensions if their dependencies are present.

src/Optimizer/Error/CannotParseJsonData.php
ediamin: Added the docblock.

src/Optimizer/Transformer/AutoExtensions.php
ediamin: Added the docblock

ediamin:
@schlessera Updated the code according to your suggestions. Please take a look.

src/Dom/Document/Filter/SelfClosingSVGElements.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * SVG elements that are self-closing.
+     *
+     * @var string[]
+     */
+    const SELF_CLOSING_TAGS = [
+        Tag::CIRCLE,
+        Tag::G,
+        Tag::PATH,
+    ];
```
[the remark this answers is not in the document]
schlessera: This was extracted into a separate issue: #405

PR #218, opened 2021-06-01
Conditionally add boilerplate error handler script depending on ESM, not transformed attribute

schlessera:
Instead of making the boilerplate error handler syntax depend on the `transformed` flag, it should depend on whether ESM is being used.

This PR lets the `RewriteAmpUrls` transformer add an `<html>` attribute `i-amphtml-module` if ESM is enabled, and then checks this attribute to decide which variant of the boilerplate error handler to use.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
           extensions: dom, iconv, json, libxml, zip
           coverage: ${{ matrix.coverage && 'pcov' || 'none' }}
           ini-values: pcov.directory=.
-          tools: composer, cs2pr
+          tools: composer
 
       - name: Get Composer Cache Directory
         id: composer-cache
@@ -44,11 +44,6 @@ jobs:
 
       - name: Install Composer dependencies
         run: composer install
-
-      # Scan the logs for failing tests and surface that information by creating annotations and log file decorations.
-      - name: Setup problem matcher to provide annotations for PHPUnit
-        # The JSON file is provided by the `shivammathur/setup-php` action. See https://github.com/shivammathur/setup-php#problem-matchers.
-        run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
 
       - name: Run tests
         if: ${{ matrix.coverage == false }}
--- a/src/Optimizer/Transformer/AmpBoilerplateErrorHandler.php
+++ b/src/Optimizer/Transformer/AmpBoilerplateErrorHandler.php
@@ -20,6 +20,15 @@ final class AmpBoilerplateErrorHandler implements Transformer
 {
 
     /**
+     * XPath query to find an AMP runtime script using ES6 modules.
+     *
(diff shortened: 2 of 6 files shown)


schlessera:
This is what I think the logic should be for the AMP boilerplate error handler syntax. The added attribute makes it possible to check for ESM support without iterating over the nodes or running a regex. It also makes it easily possible to split validator rules into ESM and no-ESM versions, just like is done with transformed/not transformed now.

@sebastianbenz, @westonruter, @honeybadgerdontcare Thoughts?

westonruter:
This makes sense to me.

westonruter:
Nevertheless, I think this may cause some more fundamental changes to be done in the validator itself. Namely, the ESM variant of `amp-onerror` has a [tag spec](https://github.com/ampproject/amphtml/blob/2c01101c767c673776e9aac791c55c0a548dd968/validator/validator-main.protoascii#L3366-L3388) which includes `enabled_by: "transformed"`. So it is specifically using this `transformed` flag.

schlessera:
Waiting for https://github.com/ampproject/amp-toolbox/pull/1237 before rebasing this.

schlessera:
I've removed the added attribute again, so this should now mirror what the NodeJS toolbox is doing, apart from one detail:

The NodeJS uses centralized configuration values, and the boilerplate error handler transformer reuses the config value from the `RewriteAmpUrls` transformer to check if ESM is "enabled", whereas the PHP version here does not use centralized configuration values and rather checks to detect if ESM is "in use". I think that the PHP version is more correct, as there might be scenarios in the NodeJS version where it will still generate the wrong error handler. For example, if you enable ESM via the configuration, but then skip loading the `RewriteAmpUrls` transformer, the NodeJS version will not use ESM but produce the ESM boilerplate error handler.

cc @SebastianBenz

.github/workflows/test.yml
```diff
@@ -45,11 +45,6 @@ jobs:
       - name: Install Composer dependencies
         run: composer install
 
-      # Scan the logs for failing tests and surface that information by creating annotations and log file decorations.
-      - name: Setup problem matcher to provide annotations for PHPUnit
-        # The JSON file is provided by the `shivammathur/setup-php` action. See https://github.com/shivammathur/setup-php#problem-matchers.
-        run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
-
```
westonruter, line 52: How is this fixing the attribute value per 9969e8ae479465b42e648870ae0d4607223799ca?

westonruter (approved):

.github/workflows/test.yml
schlessera: The PHPUnit matcher annoyed me and made looking through the PR unnecessarily difficult. Besides, it almost always just points into the `SpecTest.php` file, so it's useless in multiple ways, not just one. :)

schlessera: But I should have added it in a separate commit, I agree. I probably just didn't notice that these changes were part of that commit.

PR #221, opened 2021-06-01
Review: Add validator spec via code generation #100

westonruter:
Follow-up on #100.


schlessera:
Regarding the TODOs for "class/method not found", these are caused by the lack of support for class maps in PhpStorm. The Nette PHP Generator package uses a class map as an autoloader, and PhpStorm doesn't include that in its index.

This would be solvable by either generating a complex '.phpstorm.meta.php' file or by generating a stubs file to add to the autoloader, but I'm not sure it is worth the effort.

As the TODOs caused PHPCS failures due to their length, I've removed those for now.

schlessera:
As a side-effect of solving one of the code insight TODOs, this PR has now added magic getters for all types of entries, and all of them now include a class doc-block with all relevant properties.

bin/src/Validator/SpecGenerator/Template/CssRuleset.php
```diff
@@ -48,10 +49,10 @@ class CssRuleset
      */
     public function getId()
     {
+        // @todo Is this necessary? Why not just use the class constant?
         return static::ID;
     }
```
schlessera, line 54: I'd like to keep this one in, as everything is accessible via methods right now. However, I'll add an interface for elements that can be ID'd.

schlessera:
@westonruter I adapted to code and resolved all of your TODOs except for the three that relate to the handling of latest extension versions. I have created a new issue for these, so that we can merge this PR now: https://github.com/ampproject/amp-toolbox-php/issues/232

src/Validator/Spec/Tag/LinkRelStylesheetForFonts.php
```diff
@@ -33,33 +49,22 @@ final class LinkRelStylesheetForFonts extends Tag
...
-            [
-                SpecRule::NAME => Attribute::CROSSORIGIN,
-            ],
-            [
-                SpecRule::NAME => Attribute::HREF,
+            Attribute::ASYNC => [],
+            Attribute::CROSSORIGIN => [],
+            Attribute::HREF => [
                 SpecRule::MANDATORY => true,
                 SpecRule::VALUE_REGEX => 'https://cdn\.materialdesignicons\.com/([0-9]+\.?)+/css/materialdesignicons\.min\.css|https://cloud\.typography\.com/[0-9]*/[0-9]*/css/fonts\.css|https://fast\.fonts\.net/.*|https://fonts\.googleapis\.com/css2?\?.*|https://fonts\.googleapis\.com/icon\?.*|https://fonts\.googleapis\.com/earlyaccess/.*\.css|https://maxcdn\.bootstrapcdn\.com/font-awesome/([0-9]+\.?)+/css/font-awesome\.min\.css(\?.*)?|https://(use|pro|kit)\.fontawesome\.com/releases/v([0-9]+\.?)+/css/[0-9a-zA-Z-]+\.css|https://(use|pro|kit)\.fontawesome\.com/[0-9a-zA-Z-]+\.css|https://use\.typekit\.net/[\w\p{L}\p{N}_]+\.css',
```
westonruter, line 56: Since this regex will be sent to `preg_match()`, we should escape the '/' delimiter. See https://github.com/ampproject/amp-wp/pull/6303/commits/afd117289ee863011cb1a7dc4e096b49bd4a5f15

schlessera: I wouldn't escape this against a random delimiter.
I would suggest using the `chr(0)` delimiter in the plugin if you want to avoid `preg_quote()`.

src/Tag.php
```diff
@@ -126,7 +126,7 @@ interface Tag
     const NEXTID              = 'nextid';
     const NOBR                = 'nobr';
     const NOSCRIPT            = 'noscript';
-    const O_P                 = 'o:p';
+    const O_P                 = 'o:p'; // @todo Will this be usable at present given PHP DOM?
```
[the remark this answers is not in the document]
schlessera: I'd leave this TODO in for now. I haven't tested this yet, and I'm also not sure we'll even encounter that in the WP plugin anywhere.

src/Validator/Spec/Tag/LinkRelStylesheetForFonts.php
schlessera: I'm merging this now to get rid of another huge PR, but we can discuss this particular issue separately.

westonruter: Interesting. I wasn't aware you could use a NUL as a delimiter.

westonruter: Seems you can't: 

image

westonruter: I've opened this as #238.

schlessera: Oh, sorry, I've remembered incorrectly, it was `chr(1)` that we have been using in WP-CLI, not `chr(0)`. The `chr(1)` character is the "start of header" (`SOH`) character, intended to control the flow of streams, and shouldn't be in use or cause conflicts in HTML.

This delimiter has been used as the default delimiter in the WP-CLI `search-replace` command when you provide a regex via the CLI, and as far as I know, it has never caused any issues.

=> https://3v4l.org/piT1j
Online PHP editor _ output for piT1j - Google Chrome 2021-06-14 at 7 51 44 AM

westonruter: Excellent. Sounds good to me.

PR #231, opened 2021-06-06
Use maxBytes from Spec instead of hard-coding it

schlessera:
We had initially hard-coded the `maxBytes` spec value for enforcing the byte limit on CSS.

This PR changes the code to use the spec value directly instead.

Note that the hard-coded value was already stale, so this also fixes a bug.

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -11,7 +11,7 @@ use AmpProject\Exception\InvalidByteSequence;
 use AmpProject\Exception\MaxCssByteCountExceeded;
 use AmpProject\Optimizer\CssRule;
 use AmpProject\Tag;
-use AmpProject\Validator\Spec\CssRuleset\AmpTransformed;
+use AmpProject\Validator\Spec\CssRuleset\AmpNoTransformed;
 use AmpProject\Validator\Spec\SpecRule;
 use DOMAttr;
 use DOMComment;
@@ -2107,13 +2107,14 @@ final class Document extends DOMDocument
     /**
      * Enforce a maximum number of bytes for the CSS.
      *
-     * @param int $maxByteCount Maximum number of bytes to limit the CSS to. A negative number disables the limit.
+     * @param int|null $maxByteCount Maximum number of bytes to limit the CSS to. A negative number disables the limit.
+     *                               If null then the max bytes from AmpNoTransformed is used.
      */
     public function enforceCssMaxByteCount($maxByteCount = null)
     {
         if ($maxByteCount === null) {
             // No need to instantiate the spec here, we can just directly reference the needed constant.
-            $maxByteCount = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
+            $maxByteCount = AmpNoTransformed::SPEC[SpecRule::MAX_BYTES];
         }
 
         $this->cssMaxByteCountEnforced = $maxByteCount;
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -2,7 +2,6 @@
 
 namespace AmpProject\Dom;
 
-use AmpProject\Amp;
 use AmpProject\Attribute;
 use AmpProject\DevMode;
 use AmpProject\Dom\Document\Encoding;
@@ -12,6 +11,9 @@ use AmpProject\Exception\InvalidByteSequence;
 use AmpProject\Exception\MaxCssByteCountExceeded;
 use AmpProject\Optimizer\CssRule;
 use AmpProject\Tag;
+use AmpProject\Validator\Spec;
+use AmpProject\Validator\Spec\CssRuleset\AmpTransformed;
+use AmpProject\Validator\Spec\SpecRule;
 use DOMAttr;
 use DOMComment;
 use DOMDocument;
@@ -2108,8 +2110,13 @@ final class Document extends DOMDocument
      *
      * @param int $maxByteCount Maximum number of bytes to limit the CSS to. A negative number disables the limit.
      */
-    public function enforceCssMaxByteCount($maxByteCount = Amp::MAX_CSS_BYTE_COUNT)
+    public function enforceCssMaxByteCount($maxByteCount = null)
     {
+        if ($maxByteCount === null) {
+            // No need to instantiate the spec here, we can just directly reference the needed constant.
+            $maxByteCount = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
+        }
+
         $this->cssMaxByteCountEnforced = $maxByteCount;
     }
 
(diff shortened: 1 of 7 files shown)


src/Dom/Document.php
```diff
@@ -2108,8 +2109,13 @@ public function isCssMaxByteCountEnforced()
      *
      * @param int $maxByteCount Maximum number of bytes to limit the CSS to. A negative number disables the limit.
      */
-    public function enforceCssMaxByteCount($maxByteCount = Amp::MAX_CSS_BYTE_COUNT)
+    public function enforceCssMaxByteCount($maxByteCount = null)
     {
+        if ($maxByteCount === null) {
+            // No need to instantiate the spec here, we can just directly reference the needed constant.
+            $maxByteCount = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
```
westonruter, line 2116: This is currently being set to `112500` not the expected `75000`. So this change is needed:

```suggestion
            $maxByteCount = AmpNoTransformed::SPEC[SpecRule::MAX_BYTES];
```

src/Dom/Document.php
```diff
@@ -2108,8 +2109,13 @@ public function isCssMaxByteCountEnforced()
      *
      * @param int $maxByteCount Maximum number of bytes to limit the CSS to. A negative number disables the limit.
```
westonruter, line 2110:
```suggestion
     * @param int|null $maxByteCount Maximum number of bytes to limit the CSS to. A negative number disables the limit.
     *                               If null then the max bytes from AmpNoTransformed is used.
```

src/Dom/Document.php
```diff
@@ -12,6 +11,8 @@
 use AmpProject\Exception\MaxCssByteCountExceeded;
 use AmpProject\Optimizer\CssRule;
 use AmpProject\Tag;
+use AmpProject\Validator\Spec\CssRuleset\AmpTransformed;
```
westonruter, line 14:
```suggestion
use AmpProject\Validator\Spec\CssRuleset\AmpNoTransformed;
```

tests/Dom/DocumentTest.php
```diff
@@ -2,14 +2,15 @@
...
 
-use AmpProject\Amp;
 use AmpProject\Attribute;
 use AmpProject\Dom\Document\Option;
 use AmpProject\Exception\InvalidByteSequence;
 use AmpProject\Exception\MaxCssByteCountExceeded;
 use AmpProject\Tag;
 use AmpProject\Tests\MarkupComparison;
 use AmpProject\Tests\TestCase;
+use AmpProject\Validator\Spec\CssRuleset\AmpTransformed;
```
westonruter, line 12:
```suggestion
use AmpProject\Validator\Spec\CssRuleset\AmpNoTransformed;
```

tests/Dom/DocumentTest.php
```diff
@@ -1007,10 +1008,12 @@ public function testByteCounts($html, $expectedAmpCustom, $expectedInline)
      */
     public function testGetRemainingCssSpace()
     {
+        $maxBytes = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
```
westonruter, line 1011:
```suggestion
        $maxBytes = AmpNoTransformed::SPEC[SpecRule::MAX_BYTES];
```

tests/Dom/DocumentTest.php
```diff
@@ -1031,10 +1034,12 @@ public function testGetRemainingCssSpace()
      */
     public function testAddAmpCustomStyleWithoutLimit()
     {
+        $maxBytes = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
```
westonruter, line 1037:
```suggestion
        $maxBytes = AmpNoTransformed::SPEC[SpecRule::MAX_BYTES];
```

tests/Dom/DocumentTest.php
```diff
@@ -1056,11 +1061,13 @@ public function testAddAmpCustomStyleWithoutLimit()
      */
     public function testAddAmpCustomStyleWithLimit()
     {
+        $maxBytes = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
```
westonruter, line 1064:
```suggestion
        $maxBytes = AmpNoTransformed::SPEC[SpecRule::MAX_BYTES];
```

tests/Dom/ElementTest.php
```diff
@@ -2,11 +2,12 @@
 
 namespace AmpProject\Dom;
 
-use AmpProject\Amp;
 use AmpProject\Attribute;
 use AmpProject\Exception\MaxCssByteCountExceeded;
 use AmpProject\Tag;
 use AmpProject\Tests\TestCase;
+use AmpProject\Validator\Spec\CssRuleset\AmpTransformed;
```
westonruter, line 9:
```suggestion
use AmpProject\Validator\Spec\CssRuleset\AmpNoTransformed;
```

tests/Dom/ElementTest.php
```diff
@@ -59,10 +60,12 @@ public function testByteCounts($html, $expected)
      */
     public function testAddInlineStyleWithoutLimit()
     {
+        $maxBytes = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
```
westonruter, line 63:
```suggestion
        $maxBytes = AmpNoTransformed::SPEC[SpecRule::MAX_BYTES];
```

tests/Dom/ElementTest.php
```diff
@@ -92,11 +95,13 @@ public function testAddInlineStyleWithoutLimit()
      */
     public function testAddInlineStyleWithLimit()
     {
+        $maxBytes = AmpTransformed::SPEC[SpecRule::MAX_BYTES];
```
westonruter, line 98:
```suggestion
        $maxBytes = AmpNoTransformed::SPEC[SpecRule::MAX_BYTES];
```

westonruter (changes requested):

src/Dom/Document.php
schlessera: Yes, I used the `AmpTransformed` one because my assumption was that we are using this for the Optimizer output (which should always be transformed).

However, I realize now that this particular default is set in the `Dom\Document` class, which is not limited to the Optimizer.

So I guess using the `AmpNoTransformed` one and then adapting the limit while running the Optimizer is the way to go.

westonruter (approved):

schlessera:
I also moved the `TransformedIdentifier` to be the very first transformer now. Given that we have special validation rules for transformed VS non-transformed, I think it makes sense to set that attribute as soon as possible.

westonruter (approved):

PR #251, opened 2021-06-22
Add MinifyHtml transformer

### LINKED ISSUE #14: Transformer: Minify HTML
## Feature description

Add a new transformer that minifies HTML. The initial implementation should remain simple and do the following transformations:

- collapse whitespace (except in &lt;pre&gt;, &lt;textarea&gt;, &lt;script&gt;)
- trim whitespace within &lt;head&gt;

Related ampproject/amp-wp#4213 

---------------

## Acceptance criteria

- [ ] Generated HTML output from the Optimizer is minified
- [ ] Minification reduces whitespace to a minimum

## Implementation brief

* See https://github.com/ampproject/amp-toolbox/pull/588 & https://github.com/ampproject/amp-toolbox/commit/952b1c4e3516f8bf9f2360463f3557fb6652dcac

## QA testing instructions

1. **Verify:** AMP optimization is enabled
2. Go to the AMP version of a page
3. use "View Source..." on the page to see the generated HTML markup
4. **Verify:** The generated HTML markup is minified and only contains a minimum of whitespace

## Demo

* 

## Changelog entry

*

ediamin:
- [x] collapses whitespace (except in pre, textarea, script tags)
- [x] trims whitespace (except in pre, textarea tags)
- [x] removes comments except if they match a given pattern
- [x] minifies JSON
- [ ] ~~minifies inline amp-script using terser~~ (see #260)

closes #14


schlessera:
@ediamin Can you rebase this PR onto the latest `main` and then ensure that all the PHPCS checks regarding docs are passing?

schlessera:
@ediamin Regarding `terser`, please create a new issue about the missing JS modification, and then add the spec tests that fail because of the need for `terser` to the `TESTS_TO_SKIP` array (see https://github.com/ampproject/amp-toolbox-php/blob/0.5.2/tests/Optimizer/SpecTest.php#L36-L46) with a link to that issue.

westonruter:
Do I understand correctly that the `amp-script` minification will be skipped for now? That seems like overkill and should be something that is done by whomever is writing JS for `amp-script`.

schlessera:
@westonruter Yes, not only is it overkill, but it is also hardly feasible in PHP. While there are libraries around, I didn't find any that supports full ES6 and minifies to the degree that `terser` does, which is a huge library.

ediamin:
@schlessera I've added the amp-script failing test in `TESTS_TO_SKIP` list. Now I have only one failing test for `removes_comments` spec. The problem is, if we have any html comment before `doctype` then the `Document` class [transforms it as a `securedDoctype`](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Dom/Document.php#L1667). And after I remove the html comment, this transformation cannot be [restored by `restoreDoctypeNode` method](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Dom/Document.php#L1683). Please suggest what would be my best approach.

schlessera:
@ediamin For the secure doctype comment, I'd suggest adding a hard-coded check in the minifier to skip this particular comment. Also, please check whether there might be other comments we are using as placeholders (both in the library as well as in the plugin) that we might want to preserve. It probably makes sense to have a configuration argument that allows a consumer of the library to provide a list of patterns to preserve.

schlessera:
We already have the `commentIgnorePattern` option. Maybe extend that one to also accept an array, and then add to that array as needed?

ediamin:
@schlessera I tried to skip the secured doctype comment from removing with this in `MinifyHtml` class,
```php
    private function minifyCommentNode(DOMComment $node, $opts)
    {
        // ....

        if (preg_match('/^amp-doctype html/i', $node->data)) {
            return;
        }

        $this->nodesToRemove[] = $node;
    }
```
It prevents removing the comment node, but since there is no comment node before this doctype comment node, the `Document::restoreDoctypeNode` won't restore the secure doctype comment and hence in final result, we will get the doctype comment like this,
```
<!--amp-DOCTYPE html--><html ⚡>...
```

In order to fix this I need to change the the `HTML_RESTORE_DOCTYPE_PATTERN` in `Document` class like this,
```php
const HTML_RESTORE_DOCTYPE_PATTERN = '/(<!--amp-)(doctype)(\s+[^>]+?)(-->)/i';
```
and also need to change the replacement pattern in `restoreDoctypeNode` method
```php
    private function restoreDoctypeNode($html)
    {
        if (! $this->securedDoctype) {
            return $html;
        }

        return preg_replace(self::HTML_RESTORE_DOCTYPE_PATTERN, '<!\2\3>', $html, 1);
    }
```
Having these changes in Document class, we can get the correct `<!DOCTYPE html>` back. 

Let me know your thoughts on this. Should I implement this changes in Document class?

schlessera:
Ah, I see, the doctype is only secured because there's an element in front of it, and that element is removed by the minify transformer.

I think it is enough to make the element in front of the secured doctype optional by turning the `+` modifier into a `*` modifier:
```diff
-    const HTML_RESTORE_DOCTYPE_PATTERN             = '/(^[^<]*(?>\s*<!--[^>]*>\s*)+<)'
+    const HTML_RESTORE_DOCTYPE_PATTERN             = '/(^[^<]*(?>\s*<!--[^>]*>\s*)*<)'
                                                     . '(!--amp-)(doctype)(\s+[^>]+?)(-->)/i';
```

No other changes should be needed then.

ediamin:
@schlessera Thank you very much. I've updated the regex pattern and it's working. The PR is ready for your final review.

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+ * @property bool $inBody                Is node in the body
+ * @property bool $commentIgnorePattern  Ignore comments of the form <!-- __AAAA_BBBB___ -->.
+ *
+ * @package ampproject/amp-toolbox
+ */
+final class MinifyHtmlConfiguration extends BaseTransformerConfiguration
+{
+
+    /**
+     * Is minification enabled.
```
schlessera, line 25:
```suggestion
     * Whether minification is enabled.
```

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * @var bool
+     */
+    const MINIFIY = 'minify';
+
+    /**
+     * Is amp-script minification enabled.
```
schlessera, line 32:
```suggestion
     * Whether amp-script minification is enabled.
```

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Is collapsing the whitespace enabled.
+     *
+     * @var bool
+     */
+    const COLLAPSE_WHITESPACE = 'collapseWhitespace';
+
+    /**
+     * Should remove comments.
```
schlessera, line 53:
```suggestion
     * Whether comments should be removed.
```

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * @var bool
+     */
+    const REMOVE_COMMENTS = 'removeComments';
+
+    /**
+     * Can collapse whitespace.
```
schlessera, line 60:
```suggestion
     * Whether whitespace can be collapsed.
```

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+ */
+final class MinifyHtmlConfiguration extends BaseTransformerConfiguration
+{
+
+    /**
+     * Is minification enabled.
+     *
+     * @var bool
+     */
+    const MINIFIY = 'minify';
```
schlessera, line 29:
```suggestion
    const MINIFY = 'minify';
```

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+     * Get the associative array of allowed keys and their respective default values.
+     *
+     * The array index is the key and the array value is the key's default value.
+     *
+     * @return array Associative array of allowed keys and their respective default values.
+     */
+    protected function getAllowedKeys()
+    {
+        return [
+            self::MINIFIY                 => true,
```
schlessera, line 90:
```suggestion
            self::MINIFY                 => true,
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Apply transformations to the provided DOM document.
+     *
+     * @param Document        $document DOM document to apply the transformations to.
+     * @param ErrorCollection $errors   Collection of errors that are collected during transformation.
+     * @return void
+     */
+    public function transform(Document $document, ErrorCollection $errors)
+    {
+        if (! $this->configuration->get(MinifyHtmlConfiguration::MINIFIY)) {
```
schlessera, line 56:
```suggestion
        if (! $this->configuration->get(MinifyHtmlConfiguration::MINIFY)) {
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        /*
+         * Store nodes for later deletion to avoid changing the tree structure
+         * while iterating the DOM.
+         */
+        $this->nodesToRemove = [];
+
+        // Recursively walk through all nodes and minify if possible.
+        $this->minifyNode($document, $this->configuration->toArray());
```
schlessera, line 67: Why is the configuration turned into an array here and needlessly copied?

The configuration is already available as an object (uses less memory) in the form of a class property (does not need to be passed around from method to method).
```suggestion
        $this->minifyNode($document);
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Apply minification to a DOM node
+     *
+     * @param DOMNode $node Node to apply the transformations to.
+     * @param array   $opts Configuration options.
+     * @return void
+     */
+    private function minifyNode(DOMNode $node, $opts)
```
schlessera, line 81:
```suggestion
     * @param DOMNode $node                  Node to apply the transformations to.
     * @param bool    $canCollapseWhitespace Whether whitespace can be collapsed.
     * @param bool    $inBody                Whether the node is in the body.
     * @return DOMNode[] Array of nodes to be removed.
     */
    private function minifyNode(DOMNode $node, $canCollapseWhitespace = true, $inBody = false)
```

src/Optimizer/Transformer/MinifyHtml.php
schlessera: Going through the different minify methods, I see now that this is done because the configuration is being abused for storing state across method calls while traversing nodes.

I'd rather have the configuration stay clean and only deal with configuration values, and the state be extracted and used as actual method arguments. This would be `$canCollapseWhitespace` and `$inBody`.

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Configuration for the MinifyHtml transformer.
+ *
+ * @property bool $minify                Is minification enabled.
+ * @property bool $minifyAmpScript       Is amp-script minification enabled.
+ * @property bool $minifyJSON            Is JSON data minification enabled.
+ * @property bool $collapseWhitespace    Is collapsing the whitespace enabled.
+ * @property bool $removeComments        Should remove comments.
+ * @property bool $canCollapseWhitespace Can collapse whitespace
+ * @property bool $inBody                Is node in the body
```
schlessera, line 16: As noted further down, the `$canCollapseWhitespace` and `$inBody` values are not actually configuration values, but rather node iteration state. They need to be removed here and added as method arguments in the transformer.

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+
+        if ($node->hasChildNodes()) {
+            foreach ($node->childNodes as $childNode) {
+                $this->minifyNode($childNode, $opts);
+            }
+        }
+    }
+
+    /**
+     * Minify a Text type DOM node
```
schlessera, line 115:
```suggestion
     * Minify a text type DOM node.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+     *
+     * @param DOMText $node Text to apply the transformations to.
+     * @param array   $opts Configuration options.
+     * @return void
+     */
+    private function minifyTextNode(DOMText $node, $opts)
```
schlessera, line 121:
```suggestion
     * @param DOMText $node                  Text to apply the transformations to.
     * @param bool    $canCollapseWhitespace Whether whitespace can be collapsed.
     * @param bool    $inBody                Whether the node is in the body.
     * @return DOMNode[] Array of nodes to be removed.
     */
    private function minifyTextNode(DOMText $node, $canCollapseWhitespace = true, $inBody = false)
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Minify/remove a comment node
+     *
+     * @param DOMComment $node Comment to apply the transformations to.
+     * @param array      $opts Configuration options.
+     * @return void
+     */
+    private function minifyCommentNode(DOMComment $node, $opts)
```
schlessera, line 148:
```suggestion
     * @param DOMComment $node                  Comment to apply the transformations to.
     * @param bool       $canCollapseWhitespace Whether whitespace can be collapsed.
     * @param bool       $inBody                Whether the node is in the body.
     * @return DOMNode[] Array of nodes to be removed.
     */
    private function minifyCommentNode(DOMComment $node, $canCollapseWhitespace = true, $inBody = false)
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Minify a script node
+     *
+     * @param Element $node Element to apply the transformations to.
+     * @param array   $opts Configuration options.
+     * @return void
+     */
+    private function minifyScriptNode(Element $node, $opts)
```
schlessera, line 173:
```suggestion
     * @param Element $node                  Element to apply the transformations to.
     * @param bool    $canCollapseWhitespace Whether whitespace can be collapsed.
     * @param bool    $inBody                Whether the node is in the body.
     */
    private function minifyScriptNode(Element $node, $canCollapseWhitespace = true, $inBody = false)
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        // In case of the main $document has `securedDoctype`.
+        if (preg_match('/^amp-doctype html/i', $node->data)) {
+            return;
+        }
+
+        $this->nodesToRemove[] = $node;
+    }
+
+    /**
+     * Minify a script node
```
schlessera, line 167:
```suggestion
     * Minify a script node.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        // Remove empty nodes.
+        if (strlen($node->data) === 0) {
+            $this->nodesToRemove[] = $node;
+        }
+    }
+
+    /**
+     * Minify/remove a comment node
```
schlessera, line 142:
```suggestion
     * Minify/remove a comment node.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+     * @var TransformerConfiguration
+     */
+    private $configuration;
+
+    /**
+     * Store nodes for later deletion
+     *
+     * @var array
+     */
+    private $nodesToRemove;
```
schlessera, line 35: These can better be managed via the return value of the different `minify*()` methods, which also makes semantic sense.
```suggestion
```

src/Optimizer/Transformer/MinifyHtml.php
schlessera, line 67: The nodes to remove can be collected via the return values.
```suggestion
        // Recursively walk through all nodes and minify if possible.
        $nodesToRemove = $this->minifyNode($document, $this->configuration->toArray());
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        /*
+         * Store nodes for later deletion to avoid changing the tree structure
+         * while iterating the DOM.
+         */
+        $this->nodesToRemove = [];
+
+        // Recursively walk through all nodes and minify if possible.
+        $this->minifyNode($document, $this->configuration->toArray());
+
+        foreach ($this->nodesToRemove as $nodeToRemove) {
```
schlessera, line 69:
```suggestion
        foreach ($nodesToRemove as $nodeToRemove) {
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+     */
+    private function minifyNode(DOMNode $node, $opts)
+    {
+        if ($node instanceof DOMText) {
+            $this->minifyTextNode($node, $opts);
+        } elseif ($node instanceof DOMComment) {
+            $this->minifyCommentNode($node, $opts);
+        } elseif ($node instanceof Element && $node->tagName === 'script') {
+            $this->minifyScriptNode($node, $opts);
+        }
```
schlessera, line 89:
```suggestion
		$nodesToRemove = [];

        if ($node instanceof DOMText) {
            $nodesToRemove = array_merge($nodesToRemove, $this->minifyTextNode($node, $canColapseWhitespace, $inBody));
        } elseif ($node instanceof DOMComment) {
            $nodesToRemove = array_merge($nodesToRemove, $this->minifyCommentNode($node, $canColapseWhitespace, $inBody));
        } elseif ($node instanceof Element && $node->tagName === Tag::SCRIPT) {
            $this->minifyScriptNode($node, $canColapseWhitespace, $inBody);
        }
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+            if ($node->tagName === 'head' || $node->tagName === 'html') {
+                $opts[MinifyHtmlConfiguration::IN_BODY] = false;
+            } elseif ($node->tagName === 'body') {
+                $opts[MinifyHtmlConfiguration::IN_BODY] = true;
+            }
+        }
+
+        if ($node->hasChildNodes()) {
+            foreach ($node->childNodes as $childNode) {
+                $this->minifyNode($childNode, $opts);
```
schlessera, line 109:
```suggestion
                $nodesToRemove = array_merge($nodesToRemove, $this->minifyNode($childNode, $$canCollapseWhitespace, $inBody));
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        // Update options based on the current node.
+        if (isset($node->tagName)) {
+            if (
+                $opts[MinifyHtmlConfiguration::CAN_COLLAPSE_WHITESPACE]
+                && !$this->canCollapseWhitespace($node->tagName)
+            ) {
+                $opts[MinifyHtmlConfiguration::CAN_COLLAPSE_WHITESPACE] = false;
+            }
```
schlessera, line 98:
```suggestion
$canCollapseWhitespace = canCollapseWhitespace && $this->canCollapseWhitespace($node->tagName);
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+
+            if ($node->tagName === 'head' || $node->tagName === 'html') {
+                $opts[MinifyHtmlConfiguration::IN_BODY] = false;
+            } elseif ($node->tagName === 'body') {
+                $opts[MinifyHtmlConfiguration::IN_BODY] = true;
+            }
```
schlessera, line 104:
```suggestion
            if ($node->tagName === Tag::HEAD || $node->tagName === Tag::HTML) {
                $inBody = false;
            } elseif ($node->tagName === Tag::BODY) {
                $inBody  = true;
            }
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        if ($node->hasChildNodes()) {
+            foreach ($node->childNodes as $childNode) {
+                $this->minifyNode($childNode, $opts);
+            }
+        }
+    }
```
schlessera, line 112:
```suggestion
		return $nodesToRemove;
    }
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+     * Minify a Text type DOM node
+     *
+     * @param DOMText $node Text to apply the transformations to.
+     * @param array   $opts Configuration options.
+     * @return void
+     */
+    private function minifyTextNode(DOMText $node, $opts)
+    {
+        if (! $node->data || ! $opts[MinifyHtmlConfiguration::COLLAPSE_WHITESPACE]) {
+            return;
```
schlessera, line 124:
```suggestion
            return [];
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        if ($opts[MinifyHtmlConfiguration::CAN_COLLAPSE_WHITESPACE]) {
```
schlessera, line 127:
```suggestion
        if ($canCollapseWhitespace) {
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+            $node->data = $this->normalizeWhitespace($node->data);
+        }
+
+        if (! $opts[MinifyHtmlConfiguration::IN_BODY]) {
```
schlessera, line 131:
```suggestion
        if (! $inBody) {
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+            $node->data = trim($node->data);
+        }
+
+        // Remove empty nodes.
+        if (strlen($node->data) === 0) {
+            $this->nodesToRemove[] = $node;
+        }
```
schlessera, line 138:
```suggestion
		// Do not remove the node if it is not empty.
        if (strlen($node->data) > 0) {
            return [];
        }

        return [$node];
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        if (preg_match($opts[MinifyHtmlConfiguration::COMMENT_IGNORE_PATTERN], $node->data)) {
+            return;
+        }
+
+        // In case of the main $document has `securedDoctype`.
+        if (preg_match('/^amp-doctype html/i', $node->data)) {
+            return;
+        }
+
+        $this->nodesToRemove[] = $node;
```
schlessera, line 163:
```suggestion
        return [$node];
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+            return;
+        }
+
+        if (preg_match($opts[MinifyHtmlConfiguration::COMMENT_IGNORE_PATTERN], $node->data)) {
+            return;
+        }
+
+        // In case of the main $document has `securedDoctype`.
+        if (preg_match('/^amp-doctype html/i', $node->data)) {
+            return;
```
schlessera, line 160:
```suggestion
            return [];
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+     * @return void
+     */
+    private function minifyCommentNode(DOMComment $node, $opts)
+    {
+        if (! $node->data || ! $opts[MinifyHtmlConfiguration::REMOVE_COMMENTS]) {
+            return;
+        }
+
+        if (preg_match($opts[MinifyHtmlConfiguration::COMMENT_IGNORE_PATTERN], $node->data)) {
+            return;
```
schlessera, line 155:
```suggestion
            return [];
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        // In case of the main $document has `securedDoctype`.
```
schlessera, line 158:
```suggestion
        // In case the main $document has `securedDoctype`.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+     * Minify/remove a comment node
+     *
+     * @param DOMComment $node Comment to apply the transformations to.
+     * @param array      $opts Configuration options.
+     * @return void
+     */
+    private function minifyCommentNode(DOMComment $node, $opts)
+    {
+        if (! $node->data || ! $opts[MinifyHtmlConfiguration::REMOVE_COMMENTS]) {
+            return;
```
schlessera, line 151:
```suggestion
            return [];
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Check whether a tag is allowed to collapse whitespace
+     *
+     * @param string $tagName The allowed tag name.
+     * @return bool
+     */
+    private function canCollapseWhitespace($tagName)
+    {
+        return (
+            'script' !== $tagName && 'style' !== $tagName && 'pre' !== $tagName && 'textarea' !== $tagName
```
schlessera, line 199:
```suggestion
            Tag::SCRIPT !== $tagName && Tag::STYLE !== $tagName && Tag::PRE !== $tagName && Tag::TEXTAREA !== $tagName
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Checks if a node is JSON type
+     *
+     * @param Element $node The element node need to be checked.
+     * @return bool
+     */
+    private function isJSON(Element $node)
+    {
+        $type = $node->attributes->getNamedItem('type');
+        return isset($type->value) && ($type->value === 'application/json' || $type->value === 'application/ld+json');
```
schlessera, line 223:
```suggestion
        $type = $node->getAttribute(Attribute::TYPE);
        return $type === Attribute::TYPE_JSON || $type->value === Attribute::TYPE_LD_JSON;
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,242 @@
...
+     *
+     * @param DOMText $node The node to be minified.
+     * @return void
+     */
+    private function minifyJson(DOMText $node)
+    {
+        $decodedData = json_decode($node->data);
+
+        if (! empty($decodedData)) {
+            $data = json_encode($decodedData, JSON_HEX_AMP | JSON_HEX_TAG | JSON_UNESCAPED_SLASHES);
```
schlessera, line 237: The error state should be checked for JSON decoding and encoding.

schlessera (changes requested):
Great work so far, @ediamin!

There are a lot of minor nitpicks in the review, mostly about comments. I had hoped that the WPCS docs enforcing would highlight these, but it seems that's not (yet) the case.

The most important change I'd like you to work through is to separate configuration from internal state. The `$canCollapseWhitespace` and `$inBody` values are not configuration values, they manage the state to be kept across node iteration, therefore they should be managed as method arguments.

Another bigger change is to use the return value of the methods to pass around `$nodesToRemove`, as there is no reason to store these as a class property.

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,143 @@
...
+ * @property bool $commentIgnorePattern  Ignore comments of the form <!-- __AAAA_BBBB___ -->.
```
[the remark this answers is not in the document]
ediamin: Thank you for the suggestions. 👍

ediamin:
Hi @schlessera, I've updated the code with following notable changes

1. Fixed the comment errors.
2. Removed `canCollapseWhitespace` and `inBody` from the configuration class.
3. Use `minifyAmpScript = false` by default and throw an `InvalidConfigurationValue` exception for `minifyAmpScript = true`.
4. Use the configuration class to get a config value whenever we need, instead of using `$this->configuration->toArray()`. 
5. Remove `$nodesToRemove` class property and collect the nodes via return values.
6. Use `Tag` and `Attribute` interfaces for tags and attributes.
7. Added a new `InvalidJson` error class to handle errors during encoding and decoding JSON data.

Please review again and let me know your feedback. Thanks.

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,141 @@
...
+ * Configuration for the MinifyHtml transformer.
+ *
+ * @property bool   $minify                Whether minification is enabled.
+ * @property bool   $minifyAmpScript       Whether amp-script minification is enabled.
+ * @property bool   $minifyJSON            Whether JSON data minification is enabled.
+ * @property bool   $collapseWhitespace    Whether collapsing whitespace is enabled.
+ * @property bool   $removeComments        Whether comments should be removed.
+ * @property bool   $canCollapseWhitespace Whether whitespace can be collapsed.
+ * @property bool   $inBody                Whether the node is in the body.
+ * @property string $commentIgnorePattern  Ignore comments of the form <!-- __AAAA_BBBB___ -->.
```
schlessera, line 18: I don't think we should reference anything from Next.js in the PHP toolbox, as all Next.js sites will rather use the NodeJS implementations.

I suggest we do the following here:
- set the pattern to an empty string by default;
- check for the empty string and skip ignoring if found;
- adapt comments to not reference Next.js;
- provide the Next.js pattern in the configuration for the `SpecTest.php` file to ensure we still succeed with the comment ignoring spec test.
- 
```suggestion
 * @property string $commentIgnorePattern  Regex pattern of comments to keep.
```

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Whether comments should be removed.
+     *
+     * @var bool
+     */
+    const REMOVE_COMMENTS = 'removeComments';
+
+    /**
+     * Ignore comments of the form <!-- __AAAA_BBBB___ --> by default (used by Next.js).
```
schlessera, line 61:
```suggestion
     * Regular expression pattern of comments to keep.
```

src/Optimizer/Configuration/MinifyHtmlConfiguration.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    protected function getAllowedKeys()
+    {
+        return [
+            self::MINIFY                  => true,
+            self::MINIFY_AMP_SCRIPT       => false,
+            self::MINIFY_JSON             => true,
+            self::COLLAPSE_WHITESPACE     => true,
+            self::REMOVE_COMMENTS         => true,
+            self::COMMENT_IGNORE_PATTERN  => '/^\s*__[a-bA-Z0-9_-]+__\s*$/',
```
schlessera, line 82:
```suggestion
            self::COMMENT_IGNORE_PATTERN  => '',
```

tests/Optimizer/SpecTest.php
```diff
@@ -222,6 +229,10 @@ public function mapConfigurationData($configurationData)
                     $mappedConfiguration[RewriteAmpUrls::class][RewriteAmpUrlsConfiguration::RTV] = $value;
                     break;
 
+                // @TODO: Need to know what is the better way to handle this?
+                case '__format':
+                    break;
```
schlessera, line 234:
```suggestion
                case '__format':
                	// Defines how spec test snapshots are generated in the NodeJS toolbox and can be ignored here.
                    break;
```

tests/Optimizer/Transformer/MinifyHtmlTest.php
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * @return array[] Associative array of data arrays.
+     */
+    public function dataTransform()
+    {
+        $jsonData = [
+            'vars' => [
+                'apid'    => 'XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX',
+                'apv'     => '1.0',
+                'apn'     => 'My AMP Website',
```
schlessera, line 36:
```suggestion
                'apid' => 'XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX',
                'apv'  => '1.0',
                'apn'  => 'My AMP Website',
```

tests/Optimizer/SpecTest.php
```diff
@@ -44,6 +46,7 @@ final class SpecTest extends TestCase
 
         'PreloadHeroImage - max-hero-image-count-param'    => 'see https://github.com/ampproject/amp-toolbox-php/issues/55',
         'PreloadHeroImage - disable_via_param'             => 'see https://github.com/ampproject/amp-toolbox-php/issues/55',
+        'MinifyHtml - minifies_inline_amp-script'          => 'see https://github.com/ampproject/amp-toolbox-php/issues/260',
```
schlessera, line 49: I separate these with a blank line between transformers.
```suggestion

        'MinifyHtml - minifies_inline_amp-script' => 'see https://github.com/ampproject/amp-toolbox-php/issues/260',
```

tests/Optimizer/Configuration/MinifyHtmlConfigurationTest.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->assertEquals('/^\s*__[a-bA-Z0-9_-]+__\s*$/', $configuration->get('commentIgnorePattern'));
+        $this->assertEquals('/^\s*__[a-bA-Z0-9_-]+__\s*$/', $configuration->commentIgnorePattern);
+        $this->assertEquals(
+            [
+                'minify'                => true,
+                'minifyAmpScript'       => false,
+                'minifyJSON'            => true,
+                'collapseWhitespace'    => true,
+                'removeComments'        => true,
+                'commentIgnorePattern'  => '/^\s*__[a-bA-Z0-9_-]+__\s*$/',
```
schlessera, line 45:
```suggestion
                'commentIgnorePattern'  => '',
```

tests/Optimizer/Configuration/MinifyHtmlConfigurationTest.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->assertEquals(false, $configuration->get('minifyAmpScript'));
+        $this->assertEquals(false, $configuration->minifyAmpScript);
+        $this->assertEquals(true, $configuration->get('minifyJSON'));
+        $this->assertEquals(true, $configuration->minifyJSON);
+        $this->assertEquals(true, $configuration->get('collapseWhitespace'));
+        $this->assertEquals(true, $configuration->collapseWhitespace);
+        $this->assertEquals(true, $configuration->get('removeComments'));
+        $this->assertEquals(true, $configuration->removeComments);
+        $this->assertEquals('/^\s*__[a-bA-Z0-9_-]+__\s*$/', $configuration->get('commentIgnorePattern'));
+        $this->assertEquals('/^\s*__[a-bA-Z0-9_-]+__\s*$/', $configuration->commentIgnorePattern);
```
schlessera, line 37:
```suggestion
        $this->assertEquals('', $configuration->get('commentIgnorePattern'));
        $this->assertEquals('', $configuration->commentIgnorePattern);
```

tests/Optimizer/Transformer/MinifyHtmlTest.php
```diff
@@ -0,0 +1,116 @@
...
+     * @dataProvider dataTransform()
+     *
+     * @param string                  $source         String of source HTML.
+     * @param string                  $expectedHtml   String of expected HTML output.
+     * @param ErrorCollection|Error[] $expectedErrors Set of expected errors.
+     */
+    public function testTransform($source, $expectedHtml, $expectedErrors = [])
+    {
+        $document      = Document::fromHtml($source);
+        $configuration = new MinifyHtmlConfiguration();
```
schlessera, line 107:
```suggestion
        $configuration = new MinifyHtmlConfiguration(
        	[
        		MinifyHtmlConfiguration::COMMENT_IGNORE_PATTERN = '/^\s*__[a-bA-Z0-9_-]+__\s*$/',
        	]
        );
```

schlessera (changes requested):

schlessera:
I just finished another round of review.
Apart from many more comment fixes, there are two main things:
- the `commentIgnorePattern` references Next.js, but that is meaningless in the PHP context.
- the JSON modification was very inefficient in terms of computing within its inner loop.

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,256 @@
...
+    {
+        $this->configuration = $configuration;
+    }
+
+    /**
+     * Apply transformations to the provided DOM document.
+     *
+     * @param Document        $document DOM document to apply the transformations to.
+     * @param ErrorCollection $errors   Collection of errors that are collected during transformation.
+     * @return void
```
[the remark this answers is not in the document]
ediamin: @schlessera Since we are implementing the transform method of `Transformer` here, should I keep this return type?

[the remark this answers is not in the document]
schlessera: Ah, yes, keep it then, to stay consistent with the interface's docblock.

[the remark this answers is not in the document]
ediamin: OK, thanks.

ediamin:
@schlessera Updated the code according to your suggestions.

schlessera:
Great job, @ediamin !

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -0,0 +1,248 @@
...
+
+        $commentIgnorePattern = $this->configuration->get(MinifyHtmlConfiguration::COMMENT_IGNORE_PATTERN);
+        if (! empty($commentIgnorePattern) && preg_match($commentIgnorePattern, $node->data)) {
+            return [];
+        }
+
+        // In case the main $document has `securedDoctype`.
+        if (preg_match('/^amp-doctype html/i', $node->data)) {
+            return [];
+        }
```
westonruter, line 162: There's another case that I think may have been missed here, and that is the Mustache template workaround for HTML tables which is documented at https://amp.dev/documentation/components/amp-mustache/#tables

image

So if a comment node appears inside of a `template[type=amp-mustache]` element, and the comment text includes `{{`, then it should be ignored.

westonruter: @ediamin Maybe you could open a new issue or PR to address this point?

westonruter: Er, I just filed an issue for it: #285.

ediamin: @westonruter Yeah sure.

ediamin: I see you've already created the issue. Thank you very much @westonruter.

PR #254, opened 2021-06-23
Add WPCS for inline doc and comments

### LINKED ISSUE #184: Missing PHPCS rules that were brought over from WPCS
The code in this repository was started as part of a WordPress plugin. Therefore, we still have code rules in place that come from the WordPress Coding Standard (WPCS) and that we keep for consistency.

However, when the reliance on WPCS was removed from this repository, some of the checks were removed as well that would be needed to ensure the above consistency.

ediamin:
This PR adds the WordPress coding standards for inline documentation and comments.

Fixes #184


schlessera:
@ediamin Adding the WPCS docs ruleset worked fine, but we need to get this into a state where all the rules that are active pass without failing the tests.

I did a pass and made some changes to disable a few of the worst offenders.

Please go over the remaining failures (via `composer cs` on your local machine) and fix them as needed. For most of these, the rules actually flag a problem, so in these cases, adapt the docs/comments so that they are valid. In case you see something that just doesn't make sense for this repository, or is too much effort to fix at this point, please add a new exclusion like I did with a corresponding comment to explain why it was disabled (and maybe even pointing to a new issue).

The goal is to have the Code Style check above pass.

ediamin:
@schlessera I've updated the rules and fixed some errors. Please review and let me know your feedback.

src/Optimizer/Transformer/ReorderHead.php
```diff
@@ -51,9 +51,10 @@ final class ReorderHead implements Transformer
      */
     const AMP_RESOURCE_HINT_SRC_PATTERN = '#(^|[\b/])cdn\.ampproject\.org($|[\b/])#i';
 
-    /*
+    /**
```
schlessera, line 54: This is not a docblock, as it doesn't document a specific element, but rather provides an explanation for the entire group of properties. It is meant to be a standard multiline comment instead.

```suggestion
    /*
```

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -900,7 +900,7 @@ private function extractHeightsAttributeCss(Document $document, Element $element
         // TODO: I'm not sure why I initially added this here, it looks very intentional.
         // However, it doesn't match what the NodeJS version does, which is to add padding-top
         // to the inline style of the element.
-        // $this->customSizerStyles[$document->getElementId($element)] = '';
+        // `$this->customSizerStyles[$document->getElementId($element)] = '';`.
```
schlessera, line 903: This is supposed to be commented-out code where I was unsure whether it could be removed (hence the TODO). If PHPCS complains about commented-out code here, use `// phpcs:disable <rule-name>` to disable that rule for this specific line.

.phpcs.xml.dist
```diff
@@ -1,42 +1,68 @@
...
+        <!-- Validator classes are generated with a "DO NOT EDIT" comment that is erroneously output as a docblock. -->
+        <exclude name="Squiz.Commenting.FileComment.SpacingAfterOpen" />
+
+        <!-- Package tags are in class docblocks, not in file docblocks. -->
+        <exclude name="Squiz.Commenting.FileComment.MissingPackageTag" />
+
+        <!-- Do not enforce short descriptions to allow for short type hints. -->
+        <exclude name="Generic.Commenting.DocComment.MissingShort" />
+
+        <!-- Skipping following sniffs because of too many occurances that requires manual fixes  -->
```
schlessera, line 61:
```suggestion
        <!-- Skipping following sniffs because of too many occurrences that require manual fixes. -->
```

schlessera (changes requested):

schlessera:
Note that PHPStan also had a failure that you need to fix. You can run that specific test via `composer analyze`.

src/Optimizer/Transformer/ServerSideRendering.php
ediamin: Updated with `// phpcs:ignore`

ediamin:
@schlessera Updated the with phpstan fixes.

schlessera:
@ediamin The tests are still failing: https://github.com/ampproject/amp-toolbox-php/pull/254/checks?check_run_id=2905350478#step:7:18

schlessera (approved):

PR #262, opened 2021-06-26
Deprecate PreloadHeroImage for OptimizeHeroImages

### LINKED ISSUE #55: Adapt PreloadHeroImages to support new OptimizeHeroImage configuration arguments
The Node.JS implementation deprecated `PreloadHeroImages` in favor of a new `OptimizeHeroImages` with more sophisticated configuration arguments.

The PHP version needs to be adapted in order to support these via the imported spec tests.

See https://github.com/ampproject/amp-toolbox/pull/1133

schlessera:
This PR adds a new `OptimizeHeroImages` transformer and deprecates the `PreloadHeroImage` transformer for it.

The `PreloadHeroImage` transformer now produces a deprecation error and then forward the `transform()` call to the new `OptimizeHeroImages` transformer. This should provide a clean upgrade path for current consumers without breaking things.

Note that we still keep the constants within `PreloadHeroImage` around, as they were not marked as `private` for lack of language support at PHP 5.6+ and third-party code might thus directly reference them.

Fixes #55


ediamin (approved):
LGTM 👍

src/Optimizer/Transformer/OptimizeHeroImages.php
```diff
@@ -0,0 +1,708 @@
...
+use AmpProject\Url;
+
+/**
+ * OptimizeHeroImages - this transformer optimizes image rendering times for hero images. For hero images it will:
+ *
+ * 1. Inject a preload hint (if possible)
+ * 2. Generate an img tag enabling the browser to render the image without the AMP runtime being loaded.
+ *
+ * Hero images are either identified automatically or can be explicitly defined by adding an `data-hero` attribute to
+ * the element.
```
westonruter, line 31:
```suggestion
 * Hero images are either identified automatically or can be explicitly defined by adding an `data-hero` attribute or
 * `data-hero-candidate` to the element. Images with `data-hero` are exclusively made the heroes. If no image exists
 * with `data-hero`, then hero images will be made from any that have the `data-hero-candidate` attribute. Lastly, if
 * no image has either `data-hero` or `data-hero-candidate`, then the first non-tiny image that occurs before two
 * paragraphs is used as the hero.
```

westonruter (changes requested):

src/Optimizer/Transformer/OptimizeHeroImages.php
```diff
@@ -0,0 +1,709 @@
...
+                $node = NodeWalker::nextNode($node);
+            }
+        }
+
+        if (count($heroImages) > 0) {
+            return $heroImages;
+        }
+
+        // Since there were no hero images, use the hero image candidates instead.
+        $heroImages = array_slice($heroImageCandidates, 0, $maxHeroImageCount);
```
westonruter, line 239: Since `$heroImages` is known to be empty here, it seemed preferable to use `array_slice()` rather than a while loop. See af51a4636e0f9b77f5556d26834d270e9b7fa444.

src/Optimizer/Transformer/OptimizeHeroImages.php
```diff
@@ -0,0 +1,709 @@
...
+
+            $heroImage = $this->detectImageWithAttribute($node, Attribute::DATA_HERO);
+            if ($heroImage) {
+                $heroImages[] = $heroImage;
+            } elseif (count($heroImageCandidates) < $maxHeroImageCount) {
+                $heroImageCandidate = $this->detectImageWithAttribute($node, Attribute::DATA_HERO_CANDIDATE);
+                if ($heroImageCandidate) {
+                    $heroImageCandidates[] = $heroImageCandidate;
+                } elseif ($seenParagraphCount < 2 && ! $heroImageFallback) {
+                    $heroImageFallback = $this->detectPossibleHeroImageFallbacks($node);
```
westonruter, line 222: I simplified the logic in d85938a868f04ac8c921137951440db4d6f3218a since I found that we were collecting all the hero image fallbacks up until the maximum, but then we were only ever using the first image fallback as a hero.

src/Optimizer/Transformer/PreloadHeroImage.php
```diff
@@ -2,24 +2,15 @@
...
-use AmpProject\Optimizer\HeroImage;
-use AmpProject\Optimizer\ImageDimensions;
 use AmpProject\Optimizer\Transformer;
 use AmpProject\Optimizer\TransformerConfiguration;
-use AmpProject\RequestDestination;
-use AmpProject\Tag;
-use AmpProject\Url;
 
 /**
  * PreloadHeroImage - this transformer optimizes image rendering times for hero images. For hero images it will:
```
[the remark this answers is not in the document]
schlessera: Done with https://github.com/ampproject/amp-toolbox-php/pull/262/commits/6120bd8d8c696d414ebac734a621aecc093d6c41

src/Optimizer/Transformer/OptimizeHeroImages.php
schlessera: Completed with https://github.com/ampproject/amp-toolbox-php/pull/262/commits/a16797d4b72b2160fb1bfa6030bdc5a31c4322af

Issue #26, opened 2020-11-12
Extract different compat "hacks" from Dom\Document into separate objects


schlessera:
The `Dom\Document` class is too large and too complex right now.

There is a lot of logic that can be extracted into separate objects. This will make testing easier, churn on `Dom\Document` lower and help documentation and maintenance overall.

Ultimately, `Dom\Document` itself should contain the actual logic only for code that directly deals with controlling the underlying `DOMDocument`. For all of the different "hacks" that are currently included, like normalizing attributes, securing unparseable markup, etc..., those can easily be extracted into separate objects each and the `Dom\Document` would only be responsible of deciding "when" to do these things, but not "how" to do them.

schlessera:
The transformations that need to take place in the raw HTML string could make use of an object that splits the document into its main structure units first. This way, we can more easily do changes like adding a tag to the head or scanning the body, and performance will likely be improved, as regexes only ever process a portion of the document after the initial split.

schlessera:
To have a clean system in place for dealing with things like this, the `Dom\Document` should allow for `DocumentMiddleware` (name?) objects to be registerable.

```php
interface DocumentMiddleware
{
	/**
	 * Preprocess the HTML to be loaded into the Dom\Document.
	 *
	 * @param string $html String of HTML markup to be preprocessed.
	 * @return string Preprocessed string of HTML markup.
	public function beforeLoad($html);

	/**
	 * Preprocess the DOM to be saved into HTML.
	 *
	 * @param Document $document Document to be preprocessed before saving it into HTML.
	public function beforeSave(Document $document);
}
```

The list of registered middleware will then be processed in one direction for the `beforeLoad()` event, and processed in the opposite direction for the `beforeSave()` event.

schlessera:
This was resolved as a selection of `Dom\Document\Filter` implementations via #271.

PR #271, opened 2021-07-06
Refactor Document class using filters

ediamin:
The main goal is to extract `Dom\Document ` logics into separate related objects that we call the Middlewares. This will make testing easier, churn on Document lower and help documentation and maintenance overall.

Related to https://github.com/ampproject/amp-toolbox-php/issues/26


src/Dom/DocumentMiddleware.php
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Preprocess the DOM to be saved into HTML.
+     *
+     * @param string   $html     String of HTML markup to be preprocessed.
+     * @param Document $document Document to be preprocessed before saving it into HTML.
+     * @param DOMNode  $node     Optional. Parameter to output a subset of the document.
+     * @return string Preprocessed string of HTML markup.
+     */
+    public function beforeSave($html, Document $document, DOMNode $node = null);
```
schlessera, line 28: The interface is fuzzy, because these methods handle two different points in time each. That's why you're dealing with both THML strings and documents in both.

As I already briefly mentioned during our last meeting, I think the complete interface will need to have 4 methods:
1. `beforeLoad(string $html): string` dealing with HTML before it is parsed into a DOM;
2. `afterLoad(Document $document): void` dealing with the DOM after it has been parsed from the HTML;
3. `beforeSave(Document $document):void` dealing with the DOM before it is being rendered into HTML again;
4. `afterSave(string $html): string` dealing with the HTML after it has been rendered out of the final DOM.
(Note that most of the types are only here for communication the intent, but should be stripped as needed in the code to stick to PHP 5.6).

Given that most middleware will only use 1-2 of these methods, I think we should turn this from an interface into an abstract base class, where the 4 methods are already implemented but do nothing. Each middleware will then implement only those that it needs.

Then, when you build an implementation, you can actually put the logic into the method you're overriding. So, instead of this:
```php
public function beforeLoad($html) {
	return $this->secureSomething($html);
}
private function secureSomething($html) {
	// logic to manipulate $html.
	return $html;
}
```
we can use this instead:
```php
public function beforeLoad($html) {
	// logic to manipulate $html.
	return $html;
}
```

src/Dom/Document.php
```diff
@@ -6,14 +6,31 @@
...
+use AmpProject\Dom\Middleware\InsertMissingCharset;
+use AmpProject\Dom\Middleware\LoadHtmlWithLibXmlFlags;
+use AmpProject\Dom\Middleware\MoveInvalidNodesToBody;
+use AmpProject\Dom\Middleware\MustacheScriptTemplates;
+use AmpProject\Dom\Middleware\NormalizeDocumentStructure;
+use AmpProject\Dom\Middleware\NormalizeHtmlAttributes;
+use AmpProject\Dom\Middleware\NoscriptElements;
+use AmpProject\Dom\Middleware\RemoveHttpEquivCharset;
+use AmpProject\Dom\Middleware\SelfClosingTags;
+use AmpProject\Dom\Middleware\SvgSourceAttributeEncoding;
```
schlessera, line 25: Instead of having individual imports for each of these, you should only import `AmpProject\Dom\Middleware` here, and then use relative namespace down below:
```suggestion
use AmpProject\Dom\Middleware;
```

src/Dom/Document.php
```diff
@@ -355,6 +329,21 @@ public function __construct($version = '', $encoding = null)
...
+            SelfClosingTags::class,
+            NoscriptElements::class,
+            DoctypeNode::class,
+            MustacheScriptTemplates::class,
+            CharacterEncoding::class,
+            HttpEquivCharset::class,
+            LoadHtmlWithLibXmlFlags::class,
+            NormalizeHtmlAttributes::class,
+            RemoveHttpEquivCharset::class,
+            SvgSourceAttributeEncoding::class,
```
schlessera, line 345: To avoid manual imports for every single one of these, we can use relative namespaces:
```suggestion
            Middleware\AmpEmojiAttribute::class,
            Middleware\AmpBindAttributes::class,
            Middleware\SelfClosingTags::class,
            Middleware\NoscriptElements::class,
            Middleware\DoctypeNode::class,
            Middleware\MustacheScriptTemplates::class,
            Middleware\CharacterEncoding::class,
            Middleware\HttpEquivCharset::class,
            Middleware\LoadHtmlWithLibXmlFlags::class,
            Middleware\NormalizeHtmlAttributes::class,
            Middleware\RemoveHttpEquivCharset::class,
            Middleware\SvgSourceAttributeEncoding::class,
```

src/Dom/Document.php
```diff
@@ -480,21 +495,13 @@ private function reset()
...
-            $this->deduplicateTag(Tag::BODY);
-            $this->moveInvalidHeadNodesToBody();
-            $this->movePostBodyNodesToBody();
-            $this->convertHeadProfileToLink();
-        }
+        $this->registerMiddlewareClass(NormalizeDocumentStructure::class, 9);
+        $this->registerMiddlewareClass(InsertMissingCharset::class, 11);
+        $this->registerMiddlewareClass(DeDuplicateTag::class, 11);
+        $this->registerMiddlewareClass(MoveInvalidNodesToBody::class, 11);
+        $this->registerMiddlewareClass(HeadProfileToLink::class, 11);
```
schlessera, line 502: Why are these not part of the central middleware registration?

schlessera (changes requested):

src/Dom/Document.php
ediamin: These middlewares will only used when we use the `loadHtml` method. The middlewares registered in the constructor will be used for both the `loadHtml` and `loadHTMLFragment` methods.

schlessera: Ah, I see. I would keep these as part of the `Document` for now and not turn them into middleware, because:
- they are not fixing a bug in the DOM implementation, but rather enforcing assumptions we make about the ideal AMP HTML structure;
- they are a one-off change to the HTML and do not need restoring afterward;
- they are less of a "fix a bug", and more of a "make Dom\Document opinionated".

ediamin:
@schlessera I've updated the code according to your suggestions. Please review and let me know your feedback. Thanks.

schlessera (changes requested):
Great work, @ediamin!

I did quite a few changes now, please review everything carefully. The biggest change is a refactor away from the "Middleware" name and concept to a concept of "Filters" with segregated interfaces:

Image 2021-07-14 at 11 53 39 PM

I adapted the code to match this. You can see that everything is a tad cleaner now with no unneeded friction. I also added logic to properly instantiate these filters with their correct dependencies, so each object only receives what it actually needs.

Furthermore, I move a lot of the constants into the correct classes to get them out of the `Dom\Document`. This might be breaking changes, so we'll have to check with the plugin later on to make sure it still references the right constants everywhere.

Here's what I'd like you to work on to bring this over the finish line:

1. I left a @TODO in the code for you with additional logic that can be extracted. Mirror what I've done with the other filters in terms of interfaces, and also move the constants over.
2. I've noticed you sometimes renamed arguments from the interface to the implementation, to make them match the code. Please always use the exact argument names when implementing an interface or override a method. Starting with PHP 8.0, PHP has unfortunately decided that argument names are part of the interface, and I don't want us to get into trouble with errors triggered by mismatching argument names. Please go over interface implementations and method overrides and ensure that you stick to the original argument names.
3. Increase code coverage. Some of the new logic is not being tested directly, which has us fail the coverage thresholds.

ediamin:
@schlessera I've updated the code according your review. Please take a look.

src/Dom/Document.php
```diff
@@ -21,6 +26,9 @@
 use DOMNodeList;
```
schlessera, line 26: I can't reach it in the PR as a suggestion here, but the import for `DOMAttr` is not being used and can be removed.

src/Dom/Document/Filter/DeduplicateTag.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        $tagNames = [Tag::HEAD, Tag::BODY];
+
+        foreach ($tagNames as $tagName) {
+            $tags = $document->getElementsByTagName($tagName);
+
+            /**
+             * Main tag to keep.
+             *
+             * @var \AmpProject\Dom\Element|null $mainTag
```
schlessera, line 35: Import should be used here:
```suggestion
             * @var Element|null $mainTag
```

src/Dom/Document/Filter/DeduplicateTag.php
```diff
@@ -0,0 +1,78 @@
...
+
+            if (null === $mainTag) {
+                continue;
+            }
+
+            while ($tags->length > 1) {
+                /**
+                 * Tag to remove.
+                 *
+                 * @var \AmpProject\Dom\Element $tagToRemove
```
schlessera, line 47:
```suggestion
                 * @var Element $tagToRemove
```

src/Dom/Document/Filter/DeduplicateTag.php
```diff
@@ -0,0 +1,78 @@
...
+
+                foreach ($tagToRemove->childNodes as $childNode) {
+                    $mainTag->appendChild($childNode->parentNode->removeChild($childNode));
+                }
+
+                while ($tagToRemove->hasAttributes()) {
+                    /**
+                     * Attribute node to move over to the main tag.
+                     *
+                     * @var \DOMAttr $attribute
```
schlessera, line 59: An import should be used here:
```suggestion
                     * @var DOMAttr $attribute
```

src/Dom/Document/Filter/DeduplicateTag.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace AmpProject\Dom\Document\Filter;
+
+use AmpProject\Dom\Document;
+use AmpProject\Dom\Document\AfterLoadFilter;
+use AmpProject\Tag;
```
schlessera, line 7:
```suggestion
use AmpProject\Dom\Document;
use AmpProject\Dom\Document\AfterLoadFilter;
use AmpProject\Dom\Element;
use AmpProject\Tag;
use DOMAttr;
```

src/Dom/Document/Filter/SelfClosingTags.php
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @param string $html HTML string to adapt.
+     * @return string Adapted HTML string.
+     */
+    public function beforeLoad($html)
+    {
+        static $regexPattern = null;
+
+        if (null === $regexPattern) {
+            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(?<!\\\\)\/)?>(?!</\1>)#';
```
schlessera, line 39: There's a superfluous escape in this regex:
```suggestion
            $regexPattern = '#<(' . implode('|', Tag::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(?<!\\\\)/)?>(?!</\1>)#';
```

schlessera (changes requested):
Only minor tweaks needed, and then we can merge this huge PR.

src/Dom/Document.php
ediamin: Removed the `use DOMAttr`.

src/Dom/Document/Filter/DeduplicateTag.php
ediamin: Updated the code.

ediamin:
Thanks @schlessera. I've update the code according to your suggestions.

schlessera:
Great work, let's merge it!

westonruter:
😍

src/Dom/Document.php
```diff
@@ -55,71 +62,6 @@ final class Document extends DOMDocument
      */
     const HTML_DOCTYPE_REGEX_PATTERN = '#<!doctype\s+html[^>]+?>#si';
 
-    /**
```
westonruter, line 58: FYI: The removals here necessitate changes to the AMP plugin. See https://github.com/ampproject/amp-wp/pull/6498.

Issue #260, opened 2021-06-25
Missing JavaScript minifier


ediamin:
In order to minify HTML in PHP we need a PHP library that minifies JavaScript as well. We need to find a suitable PHP library and add that to our code.

Related to https://github.com/ampproject/amp-toolbox-php/issues/14

swissspidy:
I've used https://github.com/mck89/peast in the past to parse JS. Example formatting: https://github.com/mck89/peast/blob/master/doc/rendering.md#compact

schlessera:
Using peast is a good idea! While having the code in the AST form, we could also safely make other modifications that might be needed, like renaming variables or so.

I'd just make peast an optional dependency then and enable modification of script when peast's presence is detected.

PR #281, opened 2021-07-15
Add amp-script minification

ediamin:
This PR adds the remaining script minification logic for the `MinifyHtml` transformer. The minification is optional and requires `mck89/peast` package. To enable the minification first install the `mck89/peast` package in your project and set `minifyAmpScript = true` of the `MinifyHtmlConfiguration`. For example,

```
$document = Document::fromHtml($source);

// Set the minifyAmpScript option
$configuration = new MinifyHtmlConfiguration([MinifyHtmlConfiguration::MINIFY_AMP_SCRIPT => true]);

$transformer = new MinifyHtml($configuration);
$errors  = new ErrorCollection();
$transformer->transform($document, $errors);
$document->saveHTML();
```

Related to https://github.com/ampproject/amp-toolbox-php/pull/251

Fixes https://github.com/ampproject/amp-toolbox-php/issues/260

#### Notes on ignoring the `minifies_inline_amp-script` spec test
Even we're using Peast to minify the amp script, [the spect test](https://github.com/ampproject/amp-toolbox-php/tree/main/tests/spec/transformers/valid/MinifyHtml/minifies_inline_amp-script) is still ignored in this PR. The spec tests we are using in this repo are originated in the [amp-toolbox](https://github.com/ampproject/amp-toolbox) repository and we are using the `Terser` package in that repo for the minification. Unfortunately, Peast and Terser work in very different ways and so we cannot use the test cases from the amp-toolbox repo in amp-toolbox-php. The output from the spec test by the terser is not the same when using the Peast. After some adjustment I was able to match the output with Peast except for a semicolon. The semicolon after `'Hello World!'` in the following snippet from the spect test [input.html](https://github.com/ampproject/amp-toolbox-php/blob/main/tests/spec/transformers/valid/MinifyHtml/minifies_inline_amp-script/input.html) cannot remove by the Peast.
```
<script id="hello-world" type="text/plain" target="amp-script">
  const btn = document.querySelector('button');
  btn.addEventListener('click', () => {
    document.body.textContent = 'Hello World!';
  });
</script>
```
The main problem is for the expression statement like `document.body.textContent = 'Hello World!';` , there is no effect on the last semicolon. It doesn’t matter if we use the semicolon in input or not, Peast will always insert a semicolon at the end in output. The `Renderer` class decides whether the type of the node(in this case **ExpressionStatement**) requires semicolon insertion or not. This Renderer class has [a fixed list of node types](https://github.com/mck89/peast/blob/master/lib/Peast/Renderer.php#L38) that do not need the semicolon. Also, the parser returns a **Program** node type at first and [there is no override option](https://github.com/mck89/peast/blob/master/lib/Peast/Renderer.php#L196) to ignore adding semicolon for the whole statement either.

Then I tried to override the `Renderer` class, but that was not successful either. I cannot add the `ExpressionStatement` syntax class in `noSemicolon` list. It’ll remove the semicolon after `'Hello World'` but it also removes any semicolon where it necessary after an `ExpressionStatement`.
I’ve tried to create a similar mechanism that [we can find in terser](https://github.com/terser/terser/blob/master/lib/output.js#L416), but it failed too. Because I cannot access the before or after character in Peast to check if I need to insert a semicolon.

Another thing I noticed is that, terser by default removes any dead or unused code, but Peast doesn’t. So, considering these situations, the spec test for the amp-script minification is still ignored.


src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -245,4 +254,39 @@ private function minifyJson(DOMText $node, ErrorCollection $errors)
...
+            $errors->add(
+                MissingPackage::withMessage('mck89/peast package is required to minify inline amp-script.')
+            );
+            return;
+        }
+
+        $parser = Peast::latest($node->data, [])->parse();
+        $renderer = new Renderer();
+        $renderer->setFormatter(new Compact());
+        $node->data = $renderer->render($parser);
```
schlessera, line 290: This should be wrapped in a `try/catch` block, with the `catch` clause adding an error to the `ErrorCollection`. This way, there's a trace if a parsing error happens, but we don't fatal, and we leave the unminified piece of JS in place.

tests/Optimizer/Transformer/MinifyHtmlTest.php
```diff
@@ -5,11 +5,13 @@
 use AmpProject\Dom\Document;
 use AmpProject\Optimizer\Configuration\MinifyHtmlConfiguration;
 use AmpProject\Optimizer\Error\InvalidJson;
+use AmpProject\Optimizer\Error\MissingPackage;
```
schlessera, line 8: Again, these should use relative namespaces:
```suggestion
use AmpProject\Optimizer\Error;
```

tests/Optimizer/Transformer/MinifyHtmlTest.php
```diff
@@ -113,4 +115,46 @@ public function testTransform($source, $expectedHtml, $expectedErrors = [])
...
+        $document      = Document::fromHtml($source);
+        $configuration = new MinifyHtmlConfiguration([MinifyHtmlConfiguration::MINIFY_AMP_SCRIPT => true]);
+        $transformer   = new MinifyHtml($configuration);
+        $errors        = new ErrorCollection();
+
+        $transformer->transform($document, $errors);
+
+        if (! class_exists(Peast::class)) {
+            $expectedErrors = [
+                new MissingPackage('mck89/peast package is required to minify inline amp-script.'),
```
schlessera, line 153:
```suggestion
                new Error\MissingPackage('The optional package mck89/peast is required to minify inline amp-script.'),
```

schlessera: Note that the `InvalidJson` needs to be "relativized" as well above.

schlessera (changes requested):
Some changes needed, mostly minor tweaks.

Also, the large piece of cruft at the end of the PR because of codecov is pretty annoying during a review. Please create a separate PR to disable the inline annotations of codecov in PRs.

ediamin:
@schlessera I've updated the code and it is ready for your review. Also, I've added a note on why the spect test is still ignored in the main PR description.

src/Optimizer/Error/CannotMinifyAmpScript.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace AmpProject\Optimizer\Error;
+
+use AmpProject\Optimizer\Error;
+
+/**
+ * Optimizer error object when unable to minify the amp script.
```
schlessera, line 8:
```suggestion
 * Optimizer error object when unable to minify an amp-script element.
```

src/Optimizer/Error/CannotMinifyAmpScript.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @package ampproject/amp-toolbox
+ */
+final class CannotMinifyAmpScript implements Error
+{
+    use ErrorProperties;
+
+    /**
+     * Instantiate an CannotMinifyAmpScript object with an error message.
```
schlessera, line 17:
```suggestion
     * Instantiate a CannotMinifyAmpScript object with an error message.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -253,21 +262,64 @@ private function minifyJson(DOMText $node, ErrorCollection $errors)
...
+    private function minifyAmpScript(DOMText $node, ErrorCollection $errors)
+    {
+        if (! class_exists(Peast::class) || ! class_exists(Renderer::class) || ! class_exists(Compact::class)) {
+            $errors->add(Error\MissingPackage::withMessage(
+                'The optional package mck89/peast package is required to minify inline amp-script.'
+            ));
+            return;
+        }
+
+        // @codeCoverageIgnoreStart
```
schlessera, line 309: Why is code coverage being ignored here? You should probably be able to trigger an exception from Peast in some way, no?

tests/Optimizer/Transformer/Error/CannotMinifyAmpScriptTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace AmpProject\Optimizer\Transformer;
+
+use AmpProject\Optimizer\Error;
+use AmpProject\Optimizer\Error\CannotMinifyAmpScript;
+use AmpProject\Tests\TestCase;
+
+/**
+ * Test the CannotMinifyAmpScriptTest error class.
```
schlessera, line 10:
```suggestion
 * Test the CannotMinifyAmpScript error class.
```

tests/Optimizer/Transformer/MinifyHtmlTest.php
```diff
@@ -145,4 +147,46 @@ public function testTransform($source, $expectedHtml, $expectedErrors = [])
...
+        $transformer   = new MinifyHtml($configuration);
+        $errors        = new ErrorCollection();
+
+        $transformer->transform($document, $errors);
+
+        if (! class_exists(Peast::class)) {
+            $expectedErrors = [
+                new MissingPackage('The optional package mck89/peast package is required to minify inline amp-script.'),
+            ];
+            $this->assertSameErrors($expectedErrors, $errors);
```
schlessera, line 187: I think we should rather skip the test here if the dependency is missing. Otherwise, we can never know whether the amp-script minification has been tested at all, as the test will always succeed either way.

In general, a mechanism like this (adapt the test based on conditions) should never be used, it is preferable to skip the test or have two separate tests instead.

```suggestion
            $this->markTestSkipped('The optional package mck89/peast is needed to test inline amp-script minification.');
```

schlessera (changes requested):

src/Optimizer/Transformer/MinifyHtml.php
ediamin: @schlessera I can test this in my local by manually adding the Peast, but how can I handle this for GitHub automated testing? I cannot use `createMock` either, cause Peast is not used as the constructor param. Should I create dummy classes just for testing purpose while testing this piece of code? Please suggest. Thanks.

tests/Optimizer/Transformer/MinifyHtmlTest.php
ediamin: Got it. Thanks.

src/Optimizer/Transformer/MinifyHtml.php
schlessera: Ah, right. Leave it as is for now, then, and create an issue noting that Peast is currently not being test and that testing it will require a change to the GHA testing setup.

ediamin: Created a new issue for the GitHub action https://github.com/ampproject/amp-toolbox-php/issues/315

tests/Optimizer/Transformer/MinifyHtmlTest.php
ediamin: @schlessera I've updated the test using separate tests - one for with Peast and another for without Peast. Without the test for the `MissingPackage` error, it is hard to achieve the perfect coverage score. Please review and let me know your feedback on this.

schlessera (approved):

PR #297, opened 2021-07-30
Consume `bundles.config.extensions.json` to amend `TagWithExtensionSpec` classes with extension version meta

westonruter:
The AMP project contains extension version information in a bundles config file: https://github.com/ampproject/amphtml/blob/main/build-system/compile/bundles.config.extensions.json

This information is complementary to what is located in the validator spec. It has key information including what specific version the the `latest` version corresponds and what versions are available as Bento components.

This PR augments `TagWithExtensionSpec` to include this additional extension version information.

It still feels somewhat not-ideal that the the extension version information gets split apart into multiple `TagWithExtensionSpec` classes. For example, `amp-twitter` scripts have a `ScriptAmpTwitter` and a `ScriptAmpTwitter2` class. The former is the classic AMP component version whereas the latter is the Bento version. Both of these classes have a `LATEST_VERSION` which points to `0.1`, and the `versions` of the former include `0.1` but the later includes just `1.0`. So this may not be the most ergonomic way to store/access this information.

Feel free to take over this PR and make changes that seem best to you. Maybe we shouldn't be extending `TagWithExtensionSpec` classes in the first place and it would be better to represent this information in a separate set of classes.


schlessera (approved):

schlessera:
@westonruter:

> It still feels somewhat not-ideal that the the extension version information gets split apart into multiple TagWithExtensionSpec classes. For example, amp-twitter scripts have a ScriptAmpTwitter and a ScriptAmpTwitter2 class. The former is the classic AMP component version whereas the latter is the Bento version. Both of these classes have a LATEST_VERSION which points to 0.1, and the versions of the former include 0.1 but the later includes just 1.0. So this may not be the most ergonomic way to store/access this information.

I added the `byAggregateExtensionSpec()` retrieval now which returns an aggregate tag when multiple tags match. The information you can get via the aggregate is of course limited, but it can for example return the combined version meta.

PR #305, opened 2021-08-09
Preserve mustache tag wrapped in comment inside template element

### LINKED ISSUE #285: Preserve HTML comments containing Mustache variables in templates contained in `script[type=text/plain]`
There's a skip case that I think may have been missed here:

https://github.com/ampproject/amp-toolbox-php/blob/9e501e232909ffe77e853964e35580cd125e6e61/src/Optimizer/Transformer/MinifyHtml.php#L148-L165

Specifically, this is the Mustache template workaround for HTML tables which is documented at https://amp.dev/documentation/components/amp-mustache/#tables

image

So if a comment node appears inside of a `template[type=amp-mustache]` element, and the comment text includes `{{`, then it should be ignored.

_Originally posted by @westonruter in https://github.com/ampproject/amp-toolbox-php/pull/251#discussion_r663291869_

ediamin:
In an amp-mustache template, user may wrap the mustache tags inside HTML comment. One of the reason behind this may be to avoid the foster parenting. Hence we could have a template like this,
```
<template type="amp-mustache">
  <div id="cart">
    <table>
      <tr>
        <!-- {{#cart_items}} -->
        <td>{{name}}</td>
        <!-- {{/cart_items}} -->
      </tr>
    </table>
  </div>
</template>
```

The current MinifyHTML transformer removes the template tags inside the comments. The goal of this PR is to preserve these template tags and fix the issue.

Fixes #285.

### DIFF
--- a/src/Optimizer/Transformer/MinifyHtml.php
+++ b/src/Optimizer/Transformer/MinifyHtml.php
@@ -5,6 +5,7 @@ namespace AmpProject\Optimizer\Transformer;
 use AmpProject\Attribute;
 use AmpProject\Dom\Document;
 use AmpProject\Dom\Element;
+use AmpProject\Extension;
 use AmpProject\Optimizer\Configuration\MinifyHtmlConfiguration;
 use AmpProject\Optimizer\Error\InvalidJson;
 use AmpProject\Optimizer\ErrorCollection;
@@ -29,6 +30,13 @@ final class MinifyHtml implements Transformer
      * @var TransformerConfiguration
      */
     private $configuration;
+
+    /**
+     * Whether the nodes being minified are inside a mustache template.
+     *
+     * @var bool
+     */
+    private $isInsideMustacheTemplate = false;
 
     /**
      * Instantiate a MinifyHtml object.
@@ -98,11 +106,19 @@ final class MinifyHtml implements Transformer
         }
 
         if ($node->hasChildNodes()) {
+            if ($this->isMustacheTemplate($node)) {
+                $this->isInsideMustacheTemplate = true;
+            }
+
             foreach ($node->childNodes as $childNode) {
                 $nodesToRemove = array_merge(
                     $nodesToRemove,
                     $this->minifyNode($childNode, $errors, $canCollapseWhitespace, $inBody)
                 );
+            }
+
+            if ($this->isMustacheTemplate($node)) {
+                $this->isInsideMustacheTemplate = false;
             }
         }
 
@@ -158,6 +174,10 @@ final class MinifyHtml implements Transformer
 
         // In case the main $document has `securedDoctype`.
         if (preg_match('/^amp-doctype html/i', $node->data)) {
+            return [];
+        }
+
+        if ($this->isInsideMustacheTemplate && preg_match($this->getMustacheTagPattern(), $node->data)) {
             return [];
         }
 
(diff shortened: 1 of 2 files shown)


src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -30,6 +31,13 @@ final class MinifyHtml implements Transformer
      */
     private $configuration;
 
+    /**
+     * Whether the nodes being minified are inside a mustache template.
```
schlessera, line 35:
```suggestion
     * Whether the node(s) being minified are inside a mustache template.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -245,4 +265,59 @@ private function minifyJson(DOMText $node, ErrorCollection $errors)
             $node->data = str_replace(['\u003E', '\u003C'], ['\u003e', '\u003c'], $data);
         }
     }
+
+    /**
+     * Checks if a node is amp-mustache template.
```
schlessera, line 270:
```suggestion
     * Checks if a node is an amp-mustache template element.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -245,4 +265,59 @@ private function minifyJson(DOMText $node, ErrorCollection $errors)
...
+     *
+     * @param DOMNode $node The dom node need to be checked.
```
schlessera, line 272:
```suggestion
     * @param DOMNode $node The dom node to be checked.
```

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -245,4 +265,59 @@ private function minifyJson(DOMText $node, ErrorCollection $errors)
...
+     * @return bool Whether the checked element is an amp-mustache template.
```
schlessera, line 273:
```suggestion
     * @return bool Whether the checked node is an amp-mustache template element.
```

schlessera (changes requested):

schlessera (approved):

src/Optimizer/Transformer/MinifyHtml.php
```diff
@@ -98,12 +106,20 @@ private function minifyNode(DOMNode $node, ErrorCollection $errors, $canCollapse
...
+            }
+
             foreach ($node->childNodes as $childNode) {
                 $nodesToRemove = array_merge(
                     $nodesToRemove,
                     $this->minifyNode($childNode, $errors, $canCollapseWhitespace, $inBody)
                 );
             }
+
+            if ($this->isMustacheTemplate($node)) {
```
westonruter, line 120: Very minor point. The check could be stored in a variable to avoid running it twice. So like:

```php
$isNodeMustacheTemplate = $this->isMustacheTemplate($node);
if ($isNodeMustacheTemplate) {
    $this->isInsideMustacheTemplate = true;
}
// ...
if ($isNodeMustacheTemplate) {
    $this->isInsideMustacheTemplate = false;
}
```

westonruter: Also, should this be checking if `$this->isInsideMustacheTemplate` is not already true? While I don't think nested `template` elements are a thing right now, it might make sense to harden this by switching `isInsideMustacheTemplate` from a boolean to an integer depth count. 

So then you'd do:

```php
$isNodeMustacheTemplate = $this->isMustacheTemplate($node);
if ($isNodeMustacheTemplate) {
    $this->mustacheTemplateDepth++;
}
// ...
if ($isNodeMustacheTemplate) {
    $this->mustacheTemplateDepth--;
}
```

And then instead of checking `$this->isInsideMustacheTemplate` below, you'd check `$this->mustacheTemplateDepth > 0` (or just use its truthy value as well).

ediamin: Hi @westonruter thanks for the suggestions. I've raised [a small PR](https://github.com/ampproject/amp-toolbox-php/pull/309) including your suggesting code.

PR #307, opened 2021-08-09
Disable the Codecov annotations check

ediamin:
Codecov inline annotations are sometimes annoying during the PR review. To improve the PR review experience this PR disables the Codecov's annotation check.


schlessera (approved):

pierlon:
It may not help in this situation, but pressing `a` when reviewing a PR hides all inline annotations.

PR #319, opened 2021-08-17
Allow CSS max byte count enforcement in `TransformedIdentifier` transformer to be configured 

westonruter:
When working on looser sanitization in the AMP plugin to allow custom CSS larger than what AMP allows, I ran into an issue where an exception was being thrown when the [`AMP_Validation_Manager::update_admin_bar_item()` tries to set a `style` attribute](https://github.com/ampproject/amp-wp/blob/24923a403e6c010228b143375309336509f2eda1/includes/validation/class-amp-validation-manager.php#L1686):

```php
$small = $dom->createElement( 'small' );
$small->setAttribute( 'style', 'font-size: smaller' );
$small->appendChild( $dom->createTextNode( sprintf( '(%s)', $text ) ) );
$validate_link->appendChild( $small );
```

It was throwing a `MaxCssByteCountExceeded` error because my total CSS was over 200KB, and since there was no try/catch around the setting of this attribute it was causing the page to 500.

In the AMP plugin when CSS max size is not being enforced, I want to pass a flag to the `TransformedIdentifier` transformer to skip enforcing CSS max byte count. This PR makes that possible.

Also, this PR fixes a bug with the `Document::$inlineStyleByteCount` magic getter to make sure that any `style` attribute on the `html` element is also accounted for.

### DIFF
--- a/src/Optimizer/Configuration/TransformedIdentifierConfiguration.php
+++ b/src/Optimizer/Configuration/TransformedIdentifierConfiguration.php
@@ -3,11 +3,15 @@
 namespace AmpProject\Optimizer\Configuration;
 
 use AmpProject\Optimizer\Exception\InvalidConfigurationValue;
+use AmpProject\Validator\Spec\CssRuleset\AmpTransformed;
+use AmpProject\Validator\Spec\SpecRule;
 
 /**
  * Configuration for the TransformedIdentifier transformer.
  *
- * @property int $version Version number to use. Defaults to 0.
+ * @property int $version Version number to use. Defaults to 1.
+ * @property int|false $enforcedCssMaxByteCount The max bytes bytes to enforce on the document, or false to not enforce.
+ *                                              Defaults to max bytes for transformed spec.
  *
  * @package ampproject/amp-toolbox
  */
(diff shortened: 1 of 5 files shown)


westonruter:
@schlessera Anything else you'd like to see with this?

src/Optimizer/Configuration/TransformedIdentifierConfiguration.php
```diff
@@ -3,11 +3,15 @@
...
 use AmpProject\Optimizer\Exception\InvalidConfigurationValue;
+use AmpProject\Validator\Spec\CssRuleset\AmpTransformed;
+use AmpProject\Validator\Spec\SpecRule;
 
 /**
  * Configuration for the TransformedIdentifier transformer.
  *
- * @property int $version Version number to use. Defaults to 0.
+ * @property int $version Version number to use. Defaults to 1.
+ * @property int|false $enforcedCssMaxByteCount The max bytes bytes to enforce on the document, or false to not enforce.
```
schlessera, line 13:
```suggestion
 * @property int|false $enforcedCssMaxByteCount The max bytes count to enforce on the document, or false to not enforce.
```

schlessera (approved):

PR #322, opened 2021-08-19
Add LIBXML_SCHEMA_CREATE flag in Document filter

### LINKED ISSUE #306: Specific anchor tags in inline scripts breaks the script when optimized
Hi, 

We have a site which still uses AMP functionality but is not AMP cached anymore. The website had a very high LCP so we were trying to optimize it using AMP optimizer. We got a good result trying the node optimizer so implementing the php one now.

Everything works apart from a long inline script we have in the head tag. This script is getting closed mid way by AMP Optimizer. 

I tried using custom configuration thinking it would be one of Transformer causing this but no luck. Even with just 
`Transformer\RewriteAmpUrls::class,`

I still have the issue.

Would be have any configuration to disable this? I can try to post a sample JS if required.

Thanks very much.

ediamin:
With our current Document object the closing tags in any inline JS variable stripped out. For example in the following code,
```
<script>
  let foo = {
    bar: '<div>amp</div>',
  };
</script>
``` 
the closing tag in variable **bar** will be removed in the output.
```
<script>
  let foo = {
    bar: '<div>amp',
  };
</script>
``` 

It happens because the core `DOMDocument` class itself removes the tag and throws the warning
```
PHP Warning:  DOMDocument::loadHTML(): Unexpected end tag : div in Entity ...
```

This PR fix this issue using the the `LIBXML_SCHEMA_CREATE` Libxml option provided to the `DOMDocument::loadHTML()`.

Fixes #306

### DIFF
--- a/src/Dom/Document/Filter/LibxmlCompatibility.php
+++ b/src/Dom/Document/Filter/LibxmlCompatibility.php
@@ -62,8 +62,7 @@ final class LibxmlCompatibility implements BeforeLoadFilter, AfterLoadFilter
         }
 
         /**
-         * This flag prevents removing the closing tags used in inline
-         * JavaScript variables.
+         * This flag prevents removing the closing tags used in inline JavaScript variables.
          */
         if (defined('LIBXML_SCHEMA_CREATE')) {
             $this->options[Option::LIBXML_FLAGS] |= constant('LIBXML_SCHEMA_CREATE');
--- a/src/Dom/Document/Filter/LibxmlCompatibility.php
+++ b/src/Dom/Document/Filter/LibxmlCompatibility.php
@@ -61,9 +61,16 @@ final class LibxmlCompatibility implements BeforeLoadFilter, AfterLoadFilter
             $this->options[Option::LIBXML_FLAGS] |= constant('LIBXML_HTML_NODEFDTD');
         }
 
+        /**
+         * This flag prevents removing the closing tags used in inline
+         * JavaScript variables.
+         */
+        if (defined('LIBXML_SCHEMA_CREATE')) {
+            $this->options[Option::LIBXML_FLAGS] |= constant('LIBXML_SCHEMA_CREATE');
+        }
+
         return $html;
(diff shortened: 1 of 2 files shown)


src/Dom/Document/Filter/LibxmlCompatibility.php
```diff
@@ -61,10 +61,17 @@ public function beforeLoad($html)
             $this->options[Option::LIBXML_FLAGS] |= constant('LIBXML_HTML_NODEFDTD');
         }
 
+        /**
+         * This flag prevents removing the closing tags used in inline
+         * JavaScript variables.
```
schlessera, line 66:
```suggestion
         * This flag prevents removing the closing tags used in inline JavaScript variables.
```

schlessera (approved):

schlessera:
Great detective work, @ediamin !

src/Dom/Document/Filter/LibxmlCompatibility.php
```diff
@@ -61,10 +61,16 @@ public function beforeLoad($html)
             $this->options[Option::LIBXML_FLAGS] |= constant('LIBXML_HTML_NODEFDTD');
         }
 
+        /**
+         * This flag prevents removing the closing tags used in inline JavaScript variables.
+         */
+        if (defined('LIBXML_SCHEMA_CREATE')) {
+            $this->options[Option::LIBXML_FLAGS] |= constant('LIBXML_SCHEMA_CREATE');
```
westonruter, line 68: Wow, how did you discover that this constant had this effect? The [description of this constant](https://www.php.net/manual/en/libxml.constants.php#:~:text=libxml_schema_create) doesn't seem to indicate this behavior:

> Create default/fixed value nodes during XSD schema validation

ediamin: With the help of Google and Stack Overflow. :) A lot of people already had this issue with DOMDocument.

PR #328, opened 2021-08-25
Add randomly-ordered unit test run to GHA workflow

schlessera:
This PR adds a single PHPUnit run that is randomly ordered.

This can detect scenarios where the test makes assumptions that are only fulfilled by global state manipulated by a previous test, but are not true in the actual logic to be tested.

In case such a randomly ordered test run fails, the random seed can be found in the GHA log so that this particular ordering can be replicated locally.

Also, to make this work, it also fixes the one test that was dependent on ordering (`Dom\ElementTest::testAddAmpAction`).

/cc @swissspidy

### DIFF
--- a/tests/Dom/ElementTest.php
+++ b/tests/Dom/ElementTest.php
@@ -169,8 +169,62 @@ class ElementTest extends TestCase
      */
     public function testAddAmpAction(Element $element, $event, $action, $expected)
     {
-        $element->addAmpAction($event, $action);
-        $this->assertEquals($expected, $element->getAttribute('on'));
+        $dom    = new Document();
+        $button = $dom->createElementWithAttributes('button', []);
+
+        // Add initial action.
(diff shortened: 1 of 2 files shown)


tests/Dom/ElementTest.php
```diff
@@ -169,8 +169,62 @@ public function getAddAmpActionData()
      */
     public function testAddAmpAction(Element $element, $event, $action, $expected)
```
ediamin, line 170: @schlessera  $event, $action and $expected params are unused now. Do we need to consider refactoring the `getAddAmpActionData` dataProvider method?

schlessera: Ah, yes, good catch. I forgot to remove the old code. Resolved via https://github.com/ampproject/amp-toolbox-php/pull/328/commits/2d48728317c93e9412b96baf2b51c65fabaeefe3

ediamin (approved):

PR #335, opened 2021-08-28
Only merge CSS rules for `display:none`

### LINKED ISSUE #333: SSR for `sizes` is not producing the same results as unoptimized version or version from amp-toolbox
## Bug Description

See test case in https://github.com/ampproject/amp-toolbox-php/issues/333#issuecomment-907393025.

----

I'm having a problem with AMP-WP running in my installation. In both standard and transitional mode (with ?amp=1), it seems that AMP-WP remove the `sizes` attribute in the `<amp-img>` tag of the template. The WordPress theme that I created completely uses the AMPHTML format.

I've also read the issue #https://github.com/ampproject/amp-wp/issues/4606, but what I'm experiencing right now is in the `<amp-img>` markup I wrote directly in the template, not from the time of conversion from `` to `<amp-img>`.

One example can be found in the  markup (directly written in template, not from `the_content()` function) I use as follows:

```html
<amp-img
    data-hero 
    alt="Some Alt"
    srcset="http://xxxx/some-img-240x135.webp 240w, http://xxxx/some-img-480x270.webp 480w, http://xxxx/some-img-720x405.webp 720w, http://xxxx/some-img-960x540.webp 960w, http://xxxx/some-img-1200x675.webp 1200w"
    sizes="(max-width: 768px) 100vw, (max-width: 1023px) 64vw, (max-width: 1215px) 454px, 550px"
    src="http://xxxx/some-img-1200x675.webp"
    width="16"
    height="9"
    layout="responsive">
</amp-img>
```

For information, the value of the `srcset` attribute above is the output of the `wp_get_attachment_image_srcset()`, while the values ​​of the `sizes` attribute are entered directly into the template. Check this out: https://developer.wordpress.org/apis/handbook/responsive-images/#customizing-responsive-image-markup

Once AMP-WP is enabled, the markup changes to:

```html
<amp-img
    data-hero
    alt="Some Alt"
    srcset="http://xxxx/some-img-240x135.webp 240w, http://xxxx/some-img-480x270.webp 480w, http://xxxx/some-img-720x405.webp 720w, http://xxxx/some-img-960x540.webp 960w, http://xxxx/some-img-1200x675.webp 1200w"
    src="http://xxxx/some-img-1200x675.webp"
    width="16"
    height="9"
    layout="responsive"
    i-amphtml-ssr
    id="i-amp-2"
    class="i-amphtml-layout-responsive i-amphtml-layout-size-defined"
    i-amphtml-layout="responsive">
</amp-img>
```

Note that the `sizes` attribute disappears from the markup above and makes the layout of the image not look as expected.  On some other amp-imgs, I've also come across cases where the sizes values ​​in the template are changing from what they should be.

In another case, the `link rel="preload"` line with the `imgsizes` attribute as below, which I placed in the `<head>`, is also missing when the AMP-WP plugin is activated.

```html
<link rel="preload" as="image" imagesrcset="http://xxxx/img/logo-80.png 80w, http://xxxx/img/logo-107.png 107w, http://xxxx/img/logo-240.png 240w, http://xxxx/img/logo-480.png 480w" imagesizes="(max-width: 768px) 80px, 107px">
```

## Expected Behaviour

The plugin does not delete or change the values of the sizes attribute that has been set in the template 

As additional information, I have also tried to optimize with AMP Optimizer (node.js) on existing AMPHTML templates and did not encounter this problem.

Is this an error caused by AMP-WP or an error from my own side? 

Thank you in advance.

## Additional context

 - WordPress version: 5.8
 - Plugin version: 2.1.3
 - AMP plugin template mode: Standard and Transitional (with ?amp=1)
 - PHP version: 8.0.9
 - OS: Manjaro Linux 
 - Browser: Latest Chrome and Firefox

---------------

## Acceptance criteria

* 

## Implementation brief

* 

## QA testing instructions

* 

## Demo

* 

## Changelog entry

*

schlessera:
We have implemented CSS selector merging to save some bytes against the CSS byte limit. However, merging CSS rules causes CSS breakage in some cases, as CSS is dependent on execution order.

As an example, consider the following CSS:
```css
.a { color: green; }
.b { color: red; }
.b { color: green; }
```
Both `.a` and `.b` will appear green.

If we merge selectors that have the same properties, we end up with the following:
```css
.a, .b { color: green; }
.b { color: red; }
```
Now, `.a` will still appear green, but `.b` all of a sudden turned red.

This PR disables CSS selector merging in general, as it is not safe and it would be too complex to implement a more robust mechanism.

However, it keeps a single hard-code scenario enabled, to keep in sync with the Node.js toolbox. It merges the selectors if the only property is `display:none`. This is needed so that the synced spec tests still pass. The relevant code in the Node.js toolbox can be found here: https://github.com/ampproject/amp-toolbox/blob/3.0.0-canary.1/packages/optimizer/lib/transformers/ApplyCommonAttributes.js#L65

Resolves #333

### DIFF
--- a/src/Optimizer/CssRule.php
+++ b/src/Optimizer/CssRule.php
@@ -208,14 +208,20 @@ final class CssRule
      */
     public function canBeMerged(CssRule $that)
     {
+        // As merging rules changes the CSS ordering, it is not a safe operation to do.
+        // Therefore, we hard-code a single scenario here that we want to support (which is the most common case).
+        // This is the only merge that is being done in the Node.js amp-toolbox, so we keep this for consistency
+        // across implementations.
+
         if ($this->mediaQuery !== $that->mediaQuery) {
             return false;
         }
 
-        if (
-            count($this->properties) !== count($that->properties)
-            || array_diff($this->properties, $that->properties)
-        ) {
+        if (count($this->properties) !== 1 || count($that->properties) !== 1) {
+            return false;
+        }
+
+        if ($this->properties[0] !== 'display:none' || $that->properties[0] !== 'display:none') {
             return false;
         }
 
--- a/tests/Optimizer/CssRuleTest.php
+++ b/tests/Optimizer/CssRuleTest.php
@@ -76,33 +76,40 @@ class CssRuleTest extends TestCase
     public function dataCssRuleMerging()
     {
         return [
+            // The only property that should be merged for now is display:none.
+            // Having generic merging in place leads to CSS breakage as the order of execution the CSS moves around.
+
+            'same display:none property' => [['.class1', 'display:none'], ['.class2', 'display:none'], true, ['', ['.class1', '.class2'], ['display:none']]],
+
+            // All of the following rules are testing generic merges, but are currently hard-coded to not be mergeable.
+
             'different single property' => [['.class1', 'color:red'], ['.class2', 'color:blue'], false],
-            'same single property'      => [['.class1', 'color:red'], ['.class2', 'color:red'], true, ['', ['.class1', '.class2'], ['color:red']]],
+            'same single property'      => [['.class1', 'color:red'], ['.class2', 'color:red'], false],
 
             'different multiple properties' => [['.class1', 'color:red;background-color:green'], ['.class2', 'color:red;background-color:blue'], false],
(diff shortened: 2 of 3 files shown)


ediamin (approved):

westonruter (approved):
LGTM

PR #343, opened 2021-09-06
Add ProtectEsiTags Document filter

### LINKED ISSUE #202: Varnish <esi> tags are being stripped by DOMDocument
Hi,

We are using varnish as a reverse proxy for ESI blocks (even on AMP pages) and this library alters `<esi:include>` tags, so varnish is not able to recognize them.

we would like to be able to configure amp-toolbox-php to tolerate the presence of <esi:*> tags
cf https://docs.oracle.com/cd/B14099_19/caching.1012/b14046/esi.htm

ediamin:
The PHP DOMDocument alters the [ESI tags](https://www.w3.org/TR/esi-lang/) that use colon in the tag and remove the `esi:` part and hence breaks the system. This PR adds a new Document filter to fix this issue.

Fixes #202

### DIFF
--- a/src/Dom/Document/Filter/ProtectEsiTags.php
+++ b/src/Dom/Document/Filter/ProtectEsiTags.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace AmpProject\Dom\Document\Filter;
+
+use AmpProject\Dom\Document\AfterSaveFilter;
+use AmpProject\Dom\Document\BeforeLoadFilter;
+
+/**
+ * Protect the esi tags from being broken.
+ *
+ * @package ampproject/amp-toolbox
+ */
+final class ProtectEsiTags implements BeforeLoadFilter, AfterSaveFilter
+{
+
+    /**
+     * List of self closing ESI tags.
+     *
+     * @link https://www.w3.org/TR/esi-lang/
+     *
+     * @var string[]
+     */
+    const SELF_CLOSING_TAGS = [
+        'esi:include',
+        'esi:comment',
+    ];
+
+    /**
+     * Preprocess the HTML to be loaded into the Dom\Document.
+     *
+     * @param string $html String of HTML markup to be preprocessed.
+     * @return string Preprocessed string of HTML markup.
+     */
+    public function beforeLoad($html)
+    {
+        $selfClosingregex = '#<(' . implode('|', self::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(?<!\\\\)/)?>(?!</\1>)#';
+
+        $html = preg_replace($selfClosingregex, '<$1$2></$1>', $html);
+        $html = preg_replace('/(<esi:include.+?)(src)=/', '$1esi-src=', $html);
+        $html = preg_replace('/(<\/?)esi:/', '$1esi-', $html);
+
+        return $html;
+    }
+
+    /**
+     * Process the Dom\Document after being saved from Dom\Document.
+     *
+     * @param string $html String of HTML markup to be preprocessed.
+     * @return string Preprocessed string of HTML markup.
+     */
+    public function afterSave($html)
+    {
+        $selfClosingregex = '#</(' . implode('|', self::SELF_CLOSING_TAGS) . ')>#i';
+
+        $html = preg_replace('/(<\/?)esi-/', '$1esi:', $html);
+        $html = preg_replace('/(<esi:include.+?)(esi-src)=/', '$1src=', $html);
+        $html = preg_replace($selfClosingregex, '', $html);
+
+        return $html;
+    }
+}
--- a/src/Dom/Document/Filter/ProtectEsiTags.php
+++ b/src/Dom/Document/Filter/ProtectEsiTags.php
@@ -50,11 +50,11 @@ final class ProtectEsiTags implements BeforeLoadFilter, AfterSaveFilter
      */
     public function afterSave($html)
     {
-        $selfClosingregex = '#</(' . implode('|', self::SELF_CLOSING_TAGS) . ')>#i';
+        $selfClosingregex = '#></(' . implode('|', self::SELF_CLOSING_TAGS) . ')>#i';
 
         $html = preg_replace('/(<\/?)esi-/', '$1esi:', $html);
         $html = preg_replace('/(<esi:include.+?)(esi-src)=/', '$1src=', $html);
-        $html = preg_replace($selfClosingregex, '', $html);
+        $html = preg_replace($selfClosingregex, '/>', $html);
 
         return $html;
     }
--- a/tests/Dom/DocumentTest.php
+++ b/tests/Dom/DocumentTest.php
@@ -482,10 +482,10 @@ class DocumentTest extends TestCase
                 '<!DOCTYPE html><html>' . $head . '<body>'
                 . '    <esi:choose>'
                 . '        <esi:when test="$(logindata{name}) == null">'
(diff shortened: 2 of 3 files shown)


westonruter:
Should this be disabled by default? The vast majority of sites wouldn't be needing it. But sites that do should easily be able to add it.

ediamin:
> Should this be disabled by default? The vast majority of sites wouldn't be needing it. But sites that do should easily be able to add it.

Good point! I think we can have the filter class in our core but make it optional by not including it in the `Document` filter classes list. This way, people will be able to consume the feature later in their code.

@schlessera please share your thought on this.

src/Dom/Document/Filter/ProtectEsiTags.php
```diff
@@ -0,0 +1,61 @@
...
+     * Preprocess the HTML to be loaded into the Dom\Document.
+     *
+     * @param string $html String of HTML markup to be preprocessed.
+     * @return string Preprocessed string of HTML markup.
+     */
+    public function beforeLoad($html)
+    {
+        $selfClosingregex = '#<(' . implode('|', self::SELF_CLOSING_TAGS) . ')([^>]*?)(?>\s*(?<!\\\\)/)?>(?!</\1>)#';
+
+        $html = preg_replace($selfClosingregex, '<$1$2></$1>', $html);
```
schlessera, line 38: Given that the subsequent replaces don't act on full open&close tags, but only on partials, I don't understand what the problem was with the self-closing tags in the first place.

Were they being stripped to enable one of the subsequent transformations?

src/Dom/Document/Filter/ProtectEsiTags.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param string $html String of HTML markup to be preprocessed.
+     * @return string Preprocessed string of HTML markup.
+     */
+    public function afterSave($html)
+    {
+        $selfClosingregex = '#</(' . implode('|', self::SELF_CLOSING_TAGS) . ')>#i';
+
+        $html = preg_replace('/(<\/?)esi-/', '$1esi:', $html);
+        $html = preg_replace('/(<esi:include.+?)(esi-src)=/', '$1src=', $html);
+        $html = preg_replace($selfClosingregex, '', $html);
```
schlessera, line 57: The self-closing tags should remain in a valid state:
```suggestion
        $selfClosingregex = '#></(' . implode('|', self::SELF_CLOSING_TAGS) . ')>#i';

        $html = preg_replace('/(<\/?)esi-/', '$1esi:', $html);
        $html = preg_replace('/(<esi:include.+?)(esi-src)=/', '$1src=', $html);
        $html = preg_replace($selfClosingregex, '/>', $html);
```

tests/Dom/DocumentTest.php
```diff
@@ -467,6 +467,29 @@ public function dataDomDocument()
...
+                '<!DOCTYPE html><html>' . $head . '<body>'
+                . '    <esi:choose>'
+                . '        <esi:when test="$(logindata{name}) == null">'
+                . '            <esi:include src="/login/$(logindata{name})">'
+                . '        </esi:when>'
+                . '        <esi:otherwise>'
+                . '            <esi:include src="/login/guest.html">'
+                . '        </esi:otherwise>'
+                . '    </esi:choose>'
+                . '</body></html>',
```
schlessera, line 491: The expected result here is broken.
```suggestion
                '<!DOCTYPE html><html>' . $head . '<body>'
                . '    <esi:choose>'
                . '        <esi:when test="$(logindata{name}) == null">'
                . '            <esi:include src="/login/$(logindata{name})"/>'
                . '        </esi:when>'
                . '        <esi:otherwise>'
                . '            <esi:include src="/login/guest.html"/>'
                . '        </esi:otherwise>'
                . '    </esi:choose>'
                . '</body></html>',
```

schlessera (changes requested):

schlessera:
> Should this be disabled by default? The vast majority of sites wouldn't be needing it. But sites that do should easily be able to add it.

We have a lot of work-arounds in place that only a minority will profit from. But unless this is a huge performance overhead, I'd prefer to keep it enabled by default, rather than to have unnecessary support requests and people with frustrating hard-to-diagnose breakage from time to time.

westonruter:
> We have a lot of work-arounds in place that only a minority will profit from. But unless this is a huge performance overhead, I'd prefer to keep it enabled by default, rather than to have unnecessary support requests and people with frustrating hard-to-diagnose breakage from time to time.

To me, ESI seems like a different category of filter. If you aren't intending to use ESI tags in the first place, then they should not be passed through or else they'll cause validation errors.

kristoferbaxter:
Absolutely agree with @westonruter here, are there places we would consider esi tags valid without an edge server?

Would validation be the edge server's concern in this model?

schlessera:
> If you aren't intending to use ESI tags in the first place, then they should not be passed through or else they'll cause validation errors.

This is not about letting them pass through or not. This is only about making sure we're not corrupting the DOM when they are encountered. They were being passed through already, only the subsequent HTML was parsed in a broken way. This ensures that we can parse a document that contains ESI as well without creating garbage or fatals.

schlessera:
> Would validation be the edge server's concern in this model?

Yes, or our own validator's concern depending on setup. The above PR is really just about "not breaking during parsing".

schlessera:
> Absolutely agree with @westonruter here, are there places we would consider esi tags valid without an edge server?

The processing here will most likely happen on the origin server. So if directives for edge servers are included, I think we should keep these and not break them, unless we're going for strict valid AMP. For non-strict validation, ESI tags should be perfectly fine, I think, as we're not expecting the content to run from an AMP cache, but be distributed through _some_ CDN.

westonruter:
> Yes, or our own validator's concern depending on setup. The above PR is really just about "not breaking during parsing".

If this is the case, aren't there other examples of XML-namespaced tags that should also be accounted for? I'm thinking of `<o:p>` for example, which is [in the validator spec](https://github.com/ampproject/amphtml/blob/24c494d7b215e0d7c7b51a40b742f4b984486563/validator/validator-main.protoascii#L3564-L3569) but we don't currently handle it. That appears to be the only case of an XML-namespaced tag that the validator is expecting. Then there's also the namespaced SVG attributes, but the PHP DOM actually parses these successfully, so that's good.

There are some instances of XML namespaces being used for tags in HTML documents, however.  There's the (deprecated) XFBML, for example. These would have the same parsing problem as ESI tags, so maybe the solution should be generalized from `ProtectEsiTags` to rather be `ProtectXmlNamespacedTags`?

src/Dom/Document/Filter/ProtectEsiTags.php
ediamin: @schlessera Should I add the tags in `\AmpProject\Tag::SELF_CLOSING_TAGS` list? I just wanted to keep them separate from the original ones.

schlessera: I just tested, and the issue with the self-closing tags is that `DOMDocument` adds things like `</esi:include>` as additional tags for these.

schlessera (approved):

schlessera:
> These would have the same parsing problem as ESI tags, so maybe the solution should be generalized from ProtectEsiTags to rather be ProtectXmlNamespacedTags?

I'd keep these as separate filters. With the ESI tags, they are technically not really XML namespace tags, they are not even meant to be part of the document. They are more like Mustache tags in that they are placeholders that have to be replaced before the "actual" HTML/XML hits the browser.

PR #358, opened 2021-09-20
Turn `TagWithExtensionSpec` into an abstract class

schlessera:
This PR fixes lots of PHPStan issues.

One of the main changes is that it turns the `TagWithExtensionSpec` interface into an abstract class that extends from `Tag`. This solves a few inconsistencies and gets rid of the `ExtensionSpec` trait.


bin/generate-validator-spec.php
```diff
@@ -115,7 +115,10 @@ function fetch_json($url, $cache = false)
     }
     echo "\n";
 
-    $latest_bundle_config_url = sprintf($bundle_config_url, $latest_release['name']);
+    // There is currently a mismatch between data in the validator spec and data in the bundle config for the targeted
+    // release. Therefore, we fall back to main branch for now.
+    // Original code: "$latest_bundle_config_url = sprintf($bundle_config_url, $latest_release['name']);".
+    $latest_bundle_config_url = sprintf($bundle_config_url, 'main');
```
schlessera, line 121: @westonruter The code you've built to pick the correct release references a release that does not yet include some of the data that the validator JSON from the CDN needs. Assuming the CDN only returns actual release specs would mean the code in here produces outdated release labels...

schlessera: I've hit this issue now with the `amp-story-captions` extension which is included in the `validator.json` from the CDN. But the bundle config did not yet include it, meaning I ended up with a `LATEST_VERSION` of `null`.

PR #369, opened 2021-10-01
Optimize ESI tags replacements

schlessera:
This is a follow-up PR to #343, which I inadvertently merged to quickly.

This PR optimizes the regex replacements so that they are "queued" and executed with a single iteration of the regex engine, which should be more efficient.


ediamin (approved):
That's brilliant. Thanks.

PR #373, opened 2021-10-04
Add OptimizeViewport transformer 

### LINKED ISSUE #126: Adapt default viewport to reduce FID
If the `viewport` tag does not contain any scale and only has `width=device-width`, double-tap to zoom is disable, which means the browser does not need to add an artificial delay of 300ms or more to detect a double-tap.

This improves FID.

See https://github.com/ampproject/amp.dev/pull/5063

ediamin:
This PR adds a new transformer OptimizeViewport to adapt default viewport to reduce FID.

Closes #126


ediamin:
FYI, I've followed the logic from [ampproject/amp-wp](https://github.com/ampproject/amp-wp/blob/develop/includes/sanitizers/class-amp-meta-sanitizer.php#L181) for optimization.

schlessera (changes requested):

src/Optimizer/DefaultConfiguration.php
```diff
@@ -38,6 +38,7 @@ class DefaultConfiguration implements Configuration
         Transformer\PreloadHeroImage::class      => Configuration\PreloadHeroImageConfiguration::class,
         Transformer\RewriteAmpUrls::class        => Configuration\RewriteAmpUrlsConfiguration::class,
         Transformer\TransformedIdentifier::class => Configuration\TransformedIdentifierConfiguration::class,
+        Transformer\OptimizeViewport::class      => Configuration\OptimizeViewportConfiguration::class,
```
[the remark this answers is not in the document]
ediamin: Documented the transformer in README.md.

src/Optimizer/Transformer/OptimizeViewport.php
```diff
@@ -0,0 +1,127 @@
...
+                    static function ($ruleName) use ($parsedRules) {
+                        return $ruleName . '=' . $parsedRules[ $ruleName ];
+                    },
+                    array_keys($parsedRules)
+                )
+            );
+        }
+
+        $element = $this->createViewportElement($document, $viewport);
+        $document->head->appendChild($element);
```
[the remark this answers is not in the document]
ediamin: OK. Thanks for the tip.

src/Optimizer/Transformer/OptimizeViewport.php
```diff
@@ -0,0 +1,131 @@
...
+                $metaTag->parentNode->removeChild($metaTag);
+            }
+
+            // Remove initial-scale=1 to leave just width=device-width in order to avoid a tap delay that hurts FID.
+            if (
+                $this->configuration->get(OptimizeViewportConfiguration::REMOVE_INITIAL_SCALE_VIEWPORT_PROPERTY)
+                && isset($parsedRules[Attribute::INITIAL_SCALE])
+                && abs((float) $parsedRules[Attribute::INITIAL_SCALE] - 1.0) < 0.0001
+            ) {
+                unset($parsedRules[Attribute::INITIAL_SCALE]);
```
schlessera, line 98: The attribute is not `initial-scale`, the attribute is `viewport`. `initial-scale` is one common _value_ of the attribute `viewport`. Because of this, I suggested using `Attribute::VIEWPORT_INITIAL_SCALE` here, to match with what we already have for some other attributes where we store common _values_ for the attribute.

See https://github.com/ampproject/amp-toolbox-php/blob/0.7.0/src/Attribute.php#L985-L1136

README.md
```diff
@@ -145,6 +145,7 @@ Note that this only lets you check whether an error "category" popped up. It can
...
 | [`TransformedIdentifier`](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/TransformedIdentifier.php) | Transformer applying the transformed identifier transformations to the HTML input. This is what marks an AMP document as "already optimized", so that the AMP runtime does not need to process it anymore. |
+| [`OptimizeViewport`](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/OptimizeViewport.php) | Transformer that normalizes and optimizes the viewport meta tag. By default it'll add `<meta name="viewport" content="width=device-width">` if viewport is missing, which is the bare minimum that AMP requires. |
```
schlessera, line 148: This list is alphabetically sorted, so this entry should appear a bit higher.

schlessera (changes requested):

src/Optimizer/Transformer/OptimizeViewport.php
ediamin: There is no `VIEWPORT_INITIAL_SCALE` constant in the `Attribute` interface. Beside we are not looking for `viewport` here. `$parsedRules` is generated from `content="width=device-width,initial-scale=1"` part of the viewport meta tag.

README.md
ediamin: Updated and added after the `OptimizeHeroImages` transformer.

src/Optimizer/Transformer/OptimizeViewport.php
schlessera: You're right, attribute value would be `initial-scale=1`, not `initial-scale`. What we need is a "viewport property"...

Ok, revert that change for now, then, and let's go with what you had initially.

ediamin: The current code is actually same as before. I'm just using the constant now instead of string `initial-scale`.

schlessera: Yes, but the `initial-scale` is not an attribute, so it doesn't belong into the `Attribute` interface. It's better to use an internal constant then.

ediamin: Updated. Added the constant in `OptimizeViewport` class and used that instead of the constant from `Attribute`.

README.md
```diff
@@ -141,6 +141,7 @@ Note that this only lets you check whether an error "category" popped up. It can
...
 | [`OptimizeHeroImages`](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/OptimizeHeroImages.php) | Transformer that optimizes image rendering times for hero images by adding preload and serverside-rendered `<img>` tags when possible. Viable hero images are `<amp-img>` tags, `<amp-video>` tags with a `poster` attribute as well as `<amp-iframe>` and `<amp-video-iframe>` tags with a `placeholder` attribute. The first viable image that is encountered is used by default, but this behavior can be overridden by adding the `data-hero` attribute to a maximum of two images. The preloads only work work images that don't use `srcset`, as that is not supported as a preload in most browsers. The serverside-rendered image will not be created for `<amp-video>` tags. |
+| [`OptimizeViewport`](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/OptimizeViewport.php) | Transformer that normalizes and optimizes the viewport meta tag. By default it'll add `<meta name="viewport" content="width=device-width">` if viewport is missing, which is the bare minimum that AMP requires. |
```
schlessera, line 144: Shortened forms like `it'll` are not usually used in formal writing.
```suggestion
| [`OptimizeViewport`](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/OptimizeViewport.php) | Transformer that normalizes and optimizes the viewport meta tag. By default it will add `<meta name="viewport" content="width=device-width">` if viewport is missing, which is the bare minimum that AMP requires. |
```

schlessera (changes requested):

schlessera (approved):

PR #374, opened 2021-10-04
Fix PHP 8.1 compatibility issues

### LINKED ISSUE #362: PHP 8.1 Compatibility
Seeing a bunch of warnings currently when running tests in the Web Stories plugin against 8.1.

Here's an excerpt:

* `Return type of AmpProject\Dom\Document::saveHTML(?DOMNode $node = null) should either be compatible with DOMDocument::saveHTML(?DOMNode $node = null): string|false, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice`
* `DOMDocument::createElement(): Passing null to parameter #2 ($value) of type string is deprecated in Dom/Document.php`
* ` DOMElement::setAttribute(): Passing null to parameter #2 ($value) of type string is deprecated in Dom/Element.php`

ediamin:
The goal of this PR is to fix some error and warning notices when run in the PHP 8.1.

Fixes #362


ediamin:
@schlessera I've followed the suggestion by [swissspidy](https://github.com/swissspidy/amp-toolbox-php/pull/1). But still having some tests failure related to the encoding in `DocumentTest`. So far I've found out that in PHP 8.1 `mb_detect_encoding` works differently. Please [see this line](https://github.com/swissspidy/amp-toolbox-php/blob/main/src/Dom/Document/Filter/DocumentEncoding.php#L129) in `DocumentEncoding` class. For the failed tests the `$detectedEncoding` is `ISO-8859-15` except for PHP 8.1. In PHP 8.1 its value is `JIS`. I tried to put the `ISO-8859-15` before `JIS` in `Encoding::DETECTION_ORDER`., but this new order causes problems in other tests.

Please suggest how should I proceed.

schlessera:
Related upstream issue: https://bugs.php.net/bug.php?id=81390

It seems like the algorithm for detecting the encoding has changed in PHP 8.1. The best approach right now seems to drastically limit the list of available encodings to more closely match was PHP 8.0 and lower was doing.

In the above issue, something like this is suggested:
```php
$encodings = ['UTF-8', 'SJIS', 'GB2312',
         'ISO-8859-1', 'ISO-8859-2', 'ISO-8859-3', 'ISO-8859-4',
         'ISO-8859-5', 'ISO-8859-6', 'ISO-8859-7', 'ISO-8859-8', 'ISO-8859-9',
         'ISO-8859-10', 'ISO-8859-13', 'ISO-8859-14', 'ISO-8859-15', 'ISO-8859-16',
         'WINDOWS-1252', 'WINDOWS-1251', 'EUC-JP', 'EUC-TW', 'KOI8-R', 'BIG-5',
         'ISO-2022-KR', 'ISO-2022-JP', 'UTF-16'
];
```

However, this encoding order might cause regressions in PHP 8.0 and lower, so we'll have to check for that and maybe even split the encoding order here by PHP version.

src/Dom/Element.php
```diff
@@ -72,6 +72,9 @@ public function addInlineStyle($style)
      */
     public function setAttribute($name, $value)
     {
+        // Make sure $value is always a string and not null.
+        $value = $value ? $value : '';
```
swissspidy, line 76: If someone passes `null`, shouldn't PHP yell at them for it? Converting `null` to empty string here would hide any potential issues in consumer code.

ediamin: I've tested this scenario with [following code snippet](https://3v4l.org/k1VTU) and found out that if we pass null as the function argument, then $value will be Null not an empty string,
```php
<?php

function test($a, $b = '')
{
    var_dump($a);
    var_dump($b); // $b is Null for second function call.
}

test('without b');
test('with null', null);
```

ediamin:
@schlessera  I've updated the existing encoding order. Looks like I did some mistake when I tried to change that order before. But it is working now for all tests. FYI the suggested encoding array does not work. PHP 8.1 produces different result than its lower versions.

ediamin:
@schlessera I've updated the code and added the ReturnTypeWillChange in validator source files.

Issue #292, opened 2021-07-26
Performance issues on `inlineCss` Transformer


jeffochoa:
I wanted to start a discussion to see if is viable/posible to refactor the inlineCss transformer to improve the overall performance.

What happens is that every-time you call the `inlineCss` transformer, two external requests are made to fetch the current AMP `Runtime Version` and the CSS styles.

I'm wondering how often those two change? (Version and Css content), isn't that something that we could cache locally to having to make those two extra request each time? 

See for reference https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/AmpRuntimeCss.php#L153 

```php
private function inlineCss(Element $ampRuntimeStyle, $version)
    {
        // Use version passed in via params if available, otherwise fetch the current prod version.
        if (! empty($version)) {
            $v0CssUrl = RuntimeVersion::appendRuntimeVersion(Amp::CACHE_HOST, $version) . '/' . self::V0_CSS;
        } else {
            $v0CssUrl = self::V0_CSS_URL;

            $options  = [
                RuntimeVersion::OPTION_CANARY => $this->configuration->get(AmpRuntimeCssConfiguration::CANARY)
            ];
            $version  = (new RuntimeVersion($this->remoteRequest))->currentVersion($options);
        }

        $ampRuntimeStyle->setAttribute(Attribute::I_AMPHTML_VERSION, $version);

        $styles = $this->configuration->get(AmpRuntimeCssConfiguration::STYLES);

        if (empty($styles)) {
            $response   = $this->remoteRequest->get($v0CssUrl);
            $statusCode = $response->getStatusCode();

            if ($statusCode < 200 || $statusCode >= 300) {
                return;
            }

            $styles = $response->getBody();
        }

        $ampRuntimeStyle->textContent = $styles;
    }
    ```

schlessera:
Yes, we can absolutely improve on that. I haven't noticed this yet because our default use case is the WordPress AMP plugin, where these requests are indeed cached.

The main issue we have to figure out here is how to do caching in a framework- and environment-independent way...

schlessera:
For reference, this is how remote requests are being routed in the AMP for WordPress plugin: https://github.com/ampproject/amp-wp/blob/6f1d2c26cba313cf70dcc36bdffee6dda78cecf1/src/AmpWpPlugin.php#L225-L232

[...]

PR #414, opened 2021-11-16
Add TemporaryFileCachedRemoteGetRequest to cache cURL request

ediamin:
This PR added a temporary file cache system to cache the remote request response. This will improve the performance issue described in issue #292.

### DIFF
--- a/src/RemoteRequest/CurlRemoteGetRequest.php
+++ b/src/RemoteRequest/CurlRemoteGetRequest.php
@@ -3,6 +3,7 @@
 namespace AmpProject\RemoteRequest;
 
 use AmpProject\Exception\FailedRemoteRequest;
+use AmpProject\Exception\FailedToGetCachedResponse;
 use AmpProject\Exception\FailedToGetFromRemoteUrl;
 use AmpProject\RemoteGetRequest;
 use AmpProject\Response;
@@ -98,6 +99,13 @@ final class CurlRemoteGetRequest implements RemoteGetRequest
      */
     public function get($url, $headers = [])
     {
+        try {
+            $cachedRequest  = new TemporaryFileCachedRemoteGetRequest();
+            return $cachedRequest->get($url, $headers);
+        } catch (FailedToGetCachedResponse $error) {
+            // Failed to get the cached response, so continue to make a cURL request.
+        }
+
         $retriesLeft = $this->retries;
         do {
             $curlHandle = curl_init();
@@ -145,7 +153,11 @@ final class CurlRemoteGetRequest implements RemoteGetRequest
                 continue;
             }
 
-            return new RemoteGetRequestResponse($body, $headers, (int) $status);
+            $response = new RemoteGetRequestResponse($body, $headers, (int) $status);
+
+            $cachedRequest->cacheResponse($response);
+
+            return $response;
         } while ($retriesLeft--);
 
         // This should never be triggered, but we want to ensure we always have a typed return value,
--- a/src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
+++ b/src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace AmpProject\RemoteRequest;
+
+use AmpProject\Exception\FailedToGetCachedResponse;
+use AmpProject\RemoteGetRequest;
+use AmpProject\Response;
+use DateTimeImmutable;
+
+/**
+ * Temporarily cache remote response.
+ *
+ * @package ampproject/amp-toolbox
+ */
+final class TemporaryFileCachedRemoteGetRequest implements RemoteGetRequest
+{
+
+    /**
+     * Prefix to use to identify cached file.
+     *
+     * @var string
+     */
+    const CACHED_FILE_PREFIX = 'amp-remote-request-';
+
+    /**
+     * Cache control header directive name.
+     *
+     * @var string
+     */
+    const CACHE_CONTROL = 'Cache-Control';
+
+    /**
+     * Default cache expiration time in seconds.
+     *
+     * The number represents one month time in seconds. It will be used by default for successful requests when
+     * the 'cache-control: max-age' was not provided.
+     *
+     * @var int
+     */
+    const EXPIRY_TIME = 2592000;
+
+    /**
+     * The temporary file abspath.
+     *
+     * @var string
+     */
+    private $file;
+
+    /**
+     * Get the cached request from a temporary file.
+     *
+     * @param string $url     URL to get.
+     * @param array  $headers Optional. Associative array of headers to send with the request. Defaults to empty array.
+     * @return Response Response for the executed request.
+     * @throws FailedToGetCachedResponse If retrieving the contents from the cache failed.
+     */
+    public function get($url, $headers = [])
+    {
+        $filename = self::CACHED_FILE_PREFIX . md5($url . json_encode($headers));
+
+        $this->file = sys_get_temp_dir() . '/' . $filename;
+
+        if (! file_exists($this->file)) {
+            throw FailedToGetCachedResponse::withUrl($url);
+        }
+
+        $cachedResponse = file_get_contents($this->file);
+
+        if (false !== $cachedResponse) {
+            if (PHP_MAJOR_VERSION >= 7) {
+                $cachedResponse = unserialize($cachedResponse, [ RemoteGetRequestResponse::class ]);
+            } else {
+                // PHP 5.6 does not provide the second $options argument yet.
+                $cachedResponse = unserialize($cachedResponse);
+            }
+        }
+
+        if (! $cachedResponse instanceof RemoteGetRequestResponse || $this->isExpired($cachedResponse)) {
+            throw FailedToGetCachedResponse::withUrl($url);
+        }
+
+        return $cachedResponse;
+    }
+
+    /**
+     * Save the response in temporary file.
+     *
+     * @param RemoteGetRequestResponse $response The response that need to be cached.
+     */
+    public function cacheResponse(RemoteGetRequestResponse $response)
+    {
+        file_put_contents($this->file, serialize($response));
+    }
+
+    /**
+     * Check whether the request is expired.
+     *
+     * @param RemoteGetRequestResponse $response The response that need to be checked.
+     * @return bool
+     */
+    private function isExpired(RemoteGetRequestResponse $response)
+    {
+        $expiry = $this->getExpiryTime($response);
+
+        return new DateTimeImmutable('now') > $expiry;
+    }
+
+    /**
+     * Calculate the expiry time.
+     *
+     * @param RemoteGetRequestResponse $response The response data need to be calculated.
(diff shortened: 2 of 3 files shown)


ediamin:
@schlessera Need your suggestions to add the test cases for `TemporaryFileCachedRemoteGetRequest`. I'm not sure what would be the best way to write tests to handle the `file_get_contents` and `file_put_contents` methods.

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,156 @@
...
+use AmpProject\RemoteGetRequest;
+use AmpProject\Response;
+use DateTimeImmutable;
+
+/**
+ * Temporarily cache remote response.
+ *
+ * @package ampproject/amp-toolbox
+ */
+final class TemporaryFileCachedRemoteGetRequest implements RemoteGetRequest
```
schlessera, line 15: The `TemporaryFileCachedRemoteGetRequest` class should be implemented as a [*Decorator*](https://refactoring.guru/design-patterns/decorator/php/example), which receives a `RemoteGetRequest` implementation and wraps the caching functionality around that.

src/RemoteRequest/CurlRemoteGetRequest.php
```diff
@@ -98,6 +99,13 @@ public function __construct($sslVerify = true, $timeout = self::DEFAULT_TIMEOUT,
      */
     public function get($url, $headers = [])
     {
+        try {
+            $cachedRequest  = new TemporaryFileCachedRemoteGetRequest();
+            return $cachedRequest->get($url, $headers);
```
schlessera, line 104: This should not be instantiated within the `CurlRemoteGetRequest`. Instead, where we are now instantiating the `CurlRemoteGetRequest` as the default implementation when none was provided, we should instantiate `new TemporaryFileCachedRemoteGetRequest(new CurlRemoteGetRequest())` instead. This works if, as commented elsewhere, the `TemporaryFileCachedRemoteGetRequest` is a decorator.

The cURL implementation should not know anything about the caching implementation.

schlessera (changes requested):

schlessera:
> I'm not sure what would be the best way to write tests to handle the file_get_contents and file_put_contents methods.

Filesystem-based manipulations can be tested in an easy way with https://github.com/bovigo/vfsStream.

ediamin:
@schlessera Updated the PR with test cases. Please take a look.

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * The decorated RemoteGetRequest instance.
+     *
+     * @var RemoteGetRequest
+     */
+    private $remoteGetRequest;
+
+    /**
+     * The directory abspath where the temporary file would be saved.
```
schlessera, line 51:
```suggestion
     * Absolute path to the directory where temporary files are saved.
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @var string
+     */
+    private $file;
+
+    /**
+     * Instantiate a TemporaryFileCachedRemoteGetRequest object.
+     *
+     * @param RemoteGetRequest $remoteGetRequest The decorated RemoteGetRequest object.
+     * @param string           $directory        Optional. The directory abspath where the temp file would be saved.
```
schlessera, line 68:
```suggestion
     * @param string           $directory        Optional. Absolute path to the directory where temporary files are saved.
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,201 @@
...
+     * @var string
+     */
+    private $directory;
+
+    /**
+     * The temporary file abspath.
+     *
+     * @var string
+     */
+    private $file;
```
schlessera, line 62: This is broken logic. The same instance of `TemporaryFileCachedRemoteGetRequest` can cache multiple different responses based on the URL that is passed into `get($url)`.

Therefore, this should not be a class property, but rather an argument that is forwarded by `get()`, as opposed to the `$directory` property which is constant for the entire lifetime of the object instance.

Otherwise, if any code would check the `$file` property without passing through the `get()` method, it would be in an inconsistent state.

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * Save the response in temporary file.
+     *
+     * @param Response $response The response that need to be cached.
+     */
+    private function cacheResponse(Response $response)
+    {
+        file_put_contents($this->file, serialize($response));
```
schlessera, line 137: As discussed above, `$this->file` should not be a class property, but rather passed in as an argument, as it is not consistent for the lifetime of the object.

Alternatively, this could directly make use of the `getTemporaryFilePath($url)` method I suggested above.

tests/RemoteRequest/TemporaryFileCachedRemoteGetRequestTest.php
```diff
@@ -0,0 +1,128 @@
...
+        $cachedRequest = $this->getTemporaryFileCachedRemoteGetRequest();
+
+        $urls            = array_keys(TestMarkup::STUBBED_REMOTE_REQUESTS);
+        $url             = $urls[0];
+        $stubbedResponse = TestMarkup::STUBBED_REMOTE_REQUESTS[$url];
+
+        $filename = TemporaryFileCachedRemoteGetRequest::CACHED_FILE_PREFIX . md5($url . json_encode([]));
+        $file     = vfsStream::url(self::DIRECTORY_NAME  . DIRECTORY_SEPARATOR . $filename);
+
+        // Test that the temp file should not exists at first.
```
schlessera, line 65:
```suggestion
        // Test that the temp file does not exist at first.
```

schlessera (changes requested):

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,209 @@
...
+                }
+
+                if ('max-age' === $cache_control_setting_parts[0]) {
+                    $max_age = abs((int)$cache_control_setting_parts[1]);
+                }
+            }
+        }
+
+        return $max_age;
+    }
```
schlessera, line 208: We're using camelCasing for variable names.
```suggestion
     * @param array $cacheControlStrings Cache-Control header value.
     * @return int The max-age value of the Cache-Control header.
     */
    private function getMaxAge($cacheControlStrings)
    {
        $maxAge = -1;

        foreach ((array) $cacheControlStrings as $cacheControlString) {
            $cacheControlParts = array_map('trim', explode(',', $cacheControlString));

            foreach ($cacheControlParts as $cacheControlPart) {
                $cacheControlSettingParts = array_map('trim', explode('=', $cacheControlPart));

                if (count($cacheControlSettingParts) !== 2) {
                    continue;
                }

                if ('max-age' === $cacheControlSettingParts[0]) {
                    $maxAge = abs((int)$cacheControlSettingParts[1]);
                }
            }
        }

        return $maxAge;
    }
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,209 @@
...
+    {
+        $file = $this->getTemporaryFilePath($url);
+
+        if (! file_exists($file)) {
+            return $this->getRemoteResponse($url, $headers);
+        }
+
+        $cachedResponse = file_get_contents($file);
+
+        if (false !== $cachedResponse) {
```
schlessera, line 88: We can avoid Yoda conditions on this project:
```suggestion
        if ($cachedResponse !== false) {
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,209 @@
...
+    private function cacheResponse($url, Response $response)
+    {
+        file_put_contents($this->getTemporaryFilePath($url), serialize($response));
+    }
+
+    /**
+     * Check whether the request is expired.
+     *
+     * @param string                   $file     The absolute path of the file that contains the response data.
+     * @param RemoteGetRequestResponse $response The response that need to be checked.
```
schlessera, line 150:
```suggestion
     * @param RemoteGetRequestResponse $response The response that needs to be checked.
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,209 @@
...
+        $expiry = $this->getExpiryTime($file, $response);
+
+        return microtime(true) > $expiry->getTimestamp();
+    }
+
+    /**
+     * Calculate the expiry time.
+     *
+     * @param string                   $file     The absolute path of the file that contains the response data.
+     * @param RemoteGetRequestResponse $response The response data need to be calculated.
```
schlessera, line 164:
```suggestion
     * @param RemoteGetRequestResponse $response The response data that needs to be calculated.
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,209 @@
...
+            $response = $this->remoteGetRequest->get($url, $headers);
+            $this->cacheResponse($url, $response);
+            return $response;
+        } catch (Exception $error) {
+            throw FailedToGetCachedResponse::withUrl($url);
+        }
+    }
+
+    /**
+     * Save the response in temporary file.
```
schlessera, line 136:
```suggestion
     * Save the response in a temporary file.
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @param string   $url      The request url.
+     * @param Response $response The response that need to be cached.
```
schlessera, line 139:
```suggestion
     * @param Response $response The response that needs to be cached.
```

src/RemoteRequest/TemporaryFileCachedRemoteGetRequest.php
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        return $cachedResponse;
+    }
+
+    /**
+     * Get the absolute path of the temporary file.
+     *
+     * @param string $url The request url.
+     * @return string The absolute path of the temporary file.
```
schlessera, line 108:
```suggestion
     * @return string The absolute path to the temporary file.
```

schlessera (changes requested):

schlessera (approved):

Issue #420, opened 2021-11-21
Bug: Transformer\MinifyHtml breaks cyrillic letter 'х' followed by whitespace


Likhachev:
```php
$unoptimizedHtml = "
А а	Б б	В в	Г г	Д д	Е е	Ё ё
Ж ж	З з	И и	Й й	К к	Л л	М м
Н н	О о	П п	Р р	С с	Т т	У у
Ф ф	Х х	Ц ц	Ч ч	Ш ш	Щ щ	Ъ ъ
Ы ы	Ь ь	Э э	Ю ю	Я я";

$transformationEngine = new TransformationEngine();
$errorCollection      = new ErrorCollection; 

$optimizedHtml = $transformationEngine->optimizeHtml(
    $unoptimizedHtml, 
    $errorCollection 
);
```

Output

```
А а	Б б	В в	Г г	Д д	Е е	Ё ё
Ж ж	З з	И и	Й й	К к	Л л	М м
Н н	О о	П п	Р р	С с	Т т	У у
Ф ф	Х � Ц ц	Ч ч	Ш ш	Щ щ	Ъ ъ
Ы ы	Ь ь	Э э	Ю ю	Я я
```

Works only with space right after the letter:

`ххххх ` => `хххх�`
but
`ххххха ` => `ххххха `

The problem in `\AmpProject\Optimizer\Transformer\MinifyHtml::normalizeWhitespace`

Temporary workaround - turn off whitespace collapsing

```php
$configurationData = [
    Transformer\MinifyHtml::class => [
        Configuration\MinifyHtmlConfiguration::COLLAPSE_WHITESPACE => false,
    ],
];
```

ediamin:
@Likhachev Thanks for reporting the bug. I've found that whitespace around Cyrillic `х` has causes issue when we try to use the vertical tab `\v` in the [preg_replace function](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/MinifyHtml.php#L239). I'll post an update with possible fix.

swissspidy:
Seeing similar issues with Persian strings like `داروی جامع امام رضا + کاربرد و نحوه مصرف دارو`:

![Screenshot 2021-12-01 at 21 16 35](https://user-images.githubusercontent.com/841956/144307334-e3e3d3f8-e907-472b-9ff2-afe294b411d8.png)

Appearance is fine again when not using the `MinifyHtml` transformer.

schlessera:
[...]

PR #421, opened 2021-11-22
Remove vertical tab replacement in MinifyHtml

ediamin:
This PR fixes the issue described in https://github.com/ampproject/amp-toolbox-php/issues/420. It removes the vertical tab `\v` replacement from the MinifyHtml::normalizeWhitespace method. Though I'm not sure, what is the exact reason why the replacement was causing the issue. After done some testing, I've found that the DOMDocument class itself doesn't work if we have a vertical tab in our source. This is handled by the `LibxmlCompatibility` filter in Dom\Document. So, since we do not have any use case for vertical tab in html source yet, I removed it completly when MinifyHtml normalizes the whitespaces in normalizeWhitespace method.


schlessera:
@ediamin I don't think this is a good solution. As far as I understand, the problem is not the vertical tab, but the fact that we're using a non-multibyte-aware string replacement mechanism (`preg_replace()`).

I've already worked on solving issues like these via a string abstraction we can use. It is currently found in the validator branch, which has not yet merged: https://github.com/ampproject/amp-toolbox-php/blob/add/194-validator-engine/src/Str.php

It allows for creating regex replacements that make use of multibyte functions when they are available.

I suggest copying over the `Str` class from the validator branch in here, and use its regex mechanism instead. My guess is that the above bug will then be fixed as long as PHP is using the `mbstring` extension (which it should for Cyrillic).

schlessera:
Oh, and if you copy over the `Str` class, make sure you also copy over its corresponding test class.

ediamin:
@schlessera I've added the Str class and updated my code. But there is one test case failing in my local 
```
❯_ phpunit tests/StrTest.php                                                                            
PHPUnit 9.5.0 by Sebastian Bergmann and contributors.

.........................................F....................... 65 / 87 ( 74%)
......................                                            87 / 87 (100%)

Time: 00:00.031, Memory: 6.00 MB

There was 1 failure:

1) AmpProject\StrTest::testCasing with data set #1 ('IñtërNÂTIÔNàlizætiøn', 'iñtërnÂtiÔnàlizætiøn', 'iñtërnâtiônàlizætiøn', 'IñTëRNÂTIÔNàLIZæTIøN', 'IÑTËRNÂTIÔNÀLIZÆTIØN')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'iñtërnÂtiÔnàlizætiøn'
+'i�t�rn�ti�n�liz�ti�n'

/Users/ediamin/www/amp-toolbox-php/tests/StrTest.php:132
/Users/ediamin/.composer/vendor/phpunit/phpunit/src/TextUI/Command.php:142
/Users/ediamin/.composer/vendor/phpunit/phpunit/src/TextUI/Command.php:95

FAILURES!
Tests: 87, Assertions: 325, Failures: 1.
```

Surprisingly, it is passing in GitHub checks. What would be the possible reason it is not working on my local machine? Any idea?

ediamin:
@schlessera I found that using `setlocale(LC_CTYPE, 'C')` before `strtolower( 'IñtërNÂTIÔNàlizætiøn' )` solves the test failing problem. But I'm not sure where to put this code in Str class. FYI, the abstract TestCase already has a setLocal method, so if we want to set the locale before the test case only, we can write `$this->setLocale(LC_CTYPE, 'C');` in StrTest::testCasing. Please suggest.

schlessera:
@ediamin I'll merge this now as this fixes the original bug. Please create a new issue with the problem you've been seeing on your local machine.

PR #434, opened 2021-12-07
Add compatibility fix to alias moved classes

schlessera:
This PR adds a mechanism to enforce the loading of compatibility fixes via Composer autoloader includes and provides a compatibility fix to register class aliases for classes that have been moved.

```
        // v0.9.0 - moving HTML-based utility into a separate `Html` subnamespace.
        'AmpProject\AtRule'             => 'AmpProject\Html\AtRule',
        'AmpProject\Attribute'          => 'AmpProject\Html\Attribute',
        'AmpProject\LengthUnit'         => 'AmpProject\Html\LengthUnit',
        'AmpProject\RequestDestination' => 'AmpProject\Html\RequestDestination',
        'AmpProject\Role'               => 'AmpProject\Html\Role',
        'AmpProject\Tag'                => 'AmpProject\Html\Tag',

        // v0.9.0 - Extracting `Encoding` out of `Dom\Document`, as it is turned into AMP value object.
        'AmpProject\Dom\Document\Encoding' => 'AmpProject\Encoding',
```

### DIFF
--- a/src/CompatibilityFix/MovedClasses.php
+++ b/src/CompatibilityFix/MovedClasses.php
@@ -38,12 +38,16 @@ final class MovedClasses implements CompatibilityFix
      */
     public static function register()
     {
-        foreach (self::ALIASES as $old => $new) {
-            if (class_exists($old)) {
-                continue;
-            }
+        spl_autoload_register(
+            static function ($oldClassName) {
+                if (! array_key_exists($oldClassName, self::ALIASES)) {
+                    return false;
+                }
 
-            class_alias($new, $old, true);
-        }
+                class_alias(self::ALIASES[$oldClassName], $oldClassName, true);
+
+                return true;
+            }
+        );
     }
 }
--- a/src/CompatibilityFix/MovedClasses.php
+++ b/src/CompatibilityFix/MovedClasses.php
@@ -38,16 +38,21 @@ final class MovedClasses implements CompatibilityFix
      */
     public static function register()
     {
-        $autoloader = static function ($oldClassName) {
-            if (! array_key_exists($oldClassName, self::ALIASES)) {
-                return false;
-            }
+        spl_autoload_register(__CLASS__ . '::autoloader');
+    }
 
-            class_alias(self::ALIASES[$oldClassName], $oldClassName, true);
+    /**
+     * Autoloader to register.
+     *
+     * @param string $oldClassName Old class name that was requested to be autoloaded.
+     * @return void
+     */
+    public static function autoloader($oldClassName)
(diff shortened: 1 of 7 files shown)


ediamin (approved):
LGTM 👍

tests/CompatibilityFix/MovedClassesTest.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @covers \AmpProject\CompatibilityFix\MovedClasses
+ * @package ampproject/amp-toolbox
+ */
+class MovedClassesTest extends TestCase
+{
+
+    public function testClassAliasesExist()
+    {
+        foreach (MovedClasses::ALIASES as $old => $new) {
```
westonruter, line 18: The `$new` variable is not being used:

```suggestion
        foreach (array_keys(MovedClasses::ALIASES) as $old) {
```

src/CompatibilityFix/MovedClasses.php
```diff
@@ -0,0 +1,58 @@
...
+     * @param string $oldClassName Old class name that was requested to be autoloaded.
+     * @return void
+     */
+    public static function autoloader($oldClassName)
+    {
+        if (! array_key_exists($oldClassName, self::ALIASES)) {
+            return;
+        }
+
+        class_alias(self::ALIASES[$oldClassName], $oldClassName, true);
```
westonruter, line 56: Using the `$autoload` argument is better than checking if the `class_exists()` or `interface_exists()`?

schlessera: The `$autoload` argument is already `true` by default, I just opted to make this explicitly visible.

I'm not sure what you mean by replacing this with `class_exists()` or `interface_exists()`. If the class/interface does not yet exist and you don't autoload it here, things will just break. If it does already exist, the autoload is not being triggered.

westonruter: Ah, right.

westonruter (approved):

westonruter:
There's another back-compat incompatibility identified by @milindmore22: https://github.com/milindmore22/amp-enhancer/commit/3dc91c63c3d6d7b841fa5b4e10f3db5ccc5f539a

`Document::XPATH_MUSTACHE_TEMPLATE_ELEMENTS_QUERY` is no longer available and needs to be replaced with `MustacheScriptTemplates::XPATH_MUSTACHE_TEMPLATE_ELEMENTS_QUERY`. I guess there's nothing that can be done since there's no magic method for missing constants?

PR #442, opened 2021-12-14
Fix PSR-12 not accepting blank lines after opening brace

schlessera:
Seems like a new rule was activated with a dependency update, and most of our test files are failing PHPCS because of a blank line after the opening brace.

This fixes that.


swissspidy:
There was a change in PHPCS 3.6.2 about this: https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.6.2

PR #443, opened 2021-12-14
Harden SVG self-closing tags Dom\Document filter

### LINKED ISSUE #441: SelfClosingSVGElements issues with large SVGs
In the Web Stories plugin we deal with potentially very large SVGs, like this one:

https://github.com/google/web-stories-wp/blob/be95b3c991e2fb6209be4f81650f8d75322f76b1/packages/stickers/src/indoor-garden-oasis/floral.js (over 400K)

This causes the `preg_replace()` call in `SelfClosingSVGElements` to fail and return `null`. That is not ideal, as it will basically replace the whole existing HTML string with `null`.

**Proposed solution:** check return value of `preg_replace()` first instead of simply blindly returning. This goes for other filters too.

Second, this filter's `#<(circle|g|path)([^>]*?)(?>\s*(?<!\\)/)?>(?!.*</\1>)#` regex seems to be very inefficient. When applying it on the above SVG on regex101.com, it errors with `Catastrophic backtracking has been detected and the execution of your expression has been halted`

**Proposed solution:** make regex more performant when dealing with such large elements

schlessera:
The `Dom\Document\Filter\SelfClosingSvgTags` filter needs to be hardened to work correctly with large SVGs.

- [x] Rewrite regexes to avoid catastrophic backtracking with large SVGs.
- [x] Ensure regex results are checked for validity before being used.

Fixes #441


schlessera:
Tests are currently failing on https://github.com/ampproject/amp-toolbox/pull/1272#issuecomment-993820905

westonruter:
This may have caused a regression: https://github.com/ampproject/amp-toolbox-php/issues/512

PR #455, opened 2021-12-21
Add InvalidOptionValue exception

### LINKED ISSUE #454: Add InvalidOptionValue exception to throw exception for invalid option value
Classes/methods that take argument as option should be able to throw an exception when we try to set an invalid one. `InvalidOptionValue` will serve that purpose. For example, `AmpBindAttributes` filter should throw error when we try to set an invalid value for `Option::AMP_BIND_SYNTAX` option.

ediamin:
This PR adds a new exception class `InvalidOptionValue`. Fixes #454.

### DIFF
--- a/src/Exception/InvalidOptionValue.php
+++ b/src/Exception/InvalidOptionValue.php
@@ -2,14 +2,14 @@
 
 namespace AmpProject\Exception;
 
-use InvalidArgumentException;
+use DomainException;
 
 /**
  * Exception thrown when an invalid option value was provided.
  *
  * @package ampproject/amp-toolbox
  */
-final class InvalidOptionValue extends InvalidArgumentException implements AmpException
+final class InvalidOptionValue extends DomainException implements AmpException
 {
     /**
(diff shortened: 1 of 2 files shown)


src/Exception/InvalidOptionValue.php
```diff
@@ -0,0 +1,30 @@
...
+namespace AmpProject\Exception;
+
+use InvalidArgumentException;
+
+/**
+ * Exception thrown when an invalid option value was provided.
+ *
+ * @package ampproject/amp-toolbox
+ */
+final class InvalidOptionValue extends InvalidArgumentException implements AmpException
```
schlessera, line 12: This should be a `DomainException` instead.

schlessera (changes requested):

src/Exception/InvalidOptionValue.php
ediamin: Updated.

schlessera (approved):

PR #476, opened 2022-01-25
Add a new simplified table formatter API 

### LINKED ISSUE #453: Add simplified use of the table formatter
The current implementation of the `TableFormatter` is cumbersome to use.

It would be great to have a simplified mechanism that takes all of the tabular data, checks the max character length for each column and then adjusts the size of the column automatically. This might produce results bigger than the current screen, but is much easier to use and will be "good enough" most of the time.

ediamin:
This PR adds a new public API to the `TableFormatter` to make it easy to use. With this we will able to format table like this,

```php
$items = [
    [ 'html', 'HTML', 'Markup Language' ],
    [ 'css', 'CSS', 'Style Sheet' ],
    [ 'js', 'JavaScript', 'Programming language' ],
];

$headers = [ 'Name', 'Title', 'Description' ];

// With headers
$tableFormatter = new TableFormatter();
$table = $tableFormatter->formatTable( $items, $headers );

/* output
+------+------------+----------------------+
| Name | Title      | Description          |
+------+------------+----------------------+
| html | HTML       | Markup Language      |
| css  | CSS        | Style Sheet          |
| js   | JavaScript | Programming language |
+------+------------+----------------------+
*/

// Without headers
$tableFormatter = new TableFormatter();
$table = $tableFormatter->formatTable( $items );

/* output
+------+------------+----------------------+
| html | HTML       | Markup Language      |
| css  | CSS        | Style Sheet          |
| js   | JavaScript | Programming language |
+------+------------+----------------------+
*/
```

Closes #453 

### To-do:
- [x] Handle long texts that exceed the width of the terminal.
- [x] Add more tests.
- [x] Add docblocks.
- [x] Fix any PHPCS or PHP Stan issues.

### DIFF
--- a/src/Cli/TableFormatter.php
+++ b/src/Cli/TableFormatter.php
@@ -41,9 +41,26 @@ class TableFormatter
      */
     protected $colors;
 
+    /**
+     * Width of each column size based on the content length.
+     *
+     * @var array
+     */
     protected $tableColumnWidths = [];
 
+    /**
+     * Maximum length of the table content.
+     *
+     * @var int
+     */
     protected $maxColumnWidth = 0;
+
+    /**
+     * Whether to wrap the table with borders.
+     *
+     * @var bool
+     */
+    protected $isBorderedTable = false;
 
     /**
      * TableFormatter constructor.
@@ -135,12 +152,23 @@ class TableFormatter
         $output = '';
         for ($index = 0; $index < $maxLength; $index++) {
             foreach ($columns as $column => $width) {
+                if ($this->isBorderedTable && $column === 0) {
+                    $output .= ltrim($this->border);
+                    $width = $width - 2;
+                }
+
                 if (isset($wrapped[$column][$index])) {
                     $value = $wrapped[$column][$index];
                 } else {
                     $value = '';
                 }
+
+                if ($this->isBorderedTable && $column === $last && $width > $this->tableColumnWidths[$last]) {
+                    $width = $this->tableColumnWidths[$last];
+                }
+
                 $chunk = $this->pad($value, $width);
+
                 if (isset($colors[$column]) && $colors[$column]) {
                     $chunk = $this->colors->wrap($chunk, $colors[$column]);
                 }
@@ -149,6 +177,10 @@ class TableFormatter
                 // Add border in-between columns.
                 if ($column != $last) {
                     $output .= $this->border;
+                }
+
+                if ($this->isBorderedTable && $column === $last) {
+                    $output .= rtrim($this->border);
                 }
             }
             $output .= "\n";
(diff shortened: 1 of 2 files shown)


schlessera:
Right now, the table formatter looks for the largest row amongst all columns. This causes all the columns to be the same width. However, this wastes a lot of space and can cause the output to easily extend the screen width if you have multiple short columns and one long column, for example.

The algorithm should instead look for the largest row _per column_, making each column only as wide as it needs for its own largest element.

So, instead of this:
```
+----------------------+----------------------+----------------------+
| Name                 | Title                | Description          |
+----------------------+----------------------+----------------------+
| html                 | HTML                 | Markup Language      |
| css                  | CSS                  | Style Sheet          |
| js                   | JavaScript           | Programming language |
+----------------------+----------------------+----------------------+
```
we want this instead:
```
+------+------------+----------------------+
| Name | Title      | Description          |
+------+------------+----------------------+
| html | HTML       | Markup Language      |
| css  | CSS        | Style Sheet          |
| js   | JavaScript | Programming language |
+------+------------+----------------------+
```

ediamin:
@schlessera Updated the code.

src/Cli/TableFormatter.php
```diff
@@ -365,4 +401,160 @@ protected function wordwrap($string, $width = 75, $break = "\n", $cut = false)
...
+     * @return string Table border.
+     */
+    protected function getTableHorizontalBorder($row, $columns)
+    {
+        $tableRow = $this->getTableRow($row, $columns);
+        $tableRow = explode("\n", $tableRow);
+        $firstRow = array_shift($tableRow);
+        $firstRow = trim($firstRow);
+        $border   = preg_replace('/[^|]/', '-', $firstRow);
+        $border   = preg_replace('/[|]/', '+', $border);
```
schlessera, line 556: These should rely on `$this->border` instead of hardcoding the border character.
```suggestion
		$borderChar = preg_quote($this->border, '/');
        $border     = preg_replace("/[^{$borderChar}]/", '-', $firstRow);
        $border     = preg_replace("/[{$borderChar}]/", '+', $border);
```

src/Cli/TableFormatter.php
```diff
@@ -131,12 +152,23 @@ public function format($columns, $texts, $colors = [])
         $output = '';
         for ($index = 0; $index < $maxLength; $index++) {
             foreach ($columns as $column => $width) {
+                if ($this->isBorderedTable && $column === 0) {
+                    $output .= ltrim($this->border);
+                    $width = $width - 2;
```
schlessera, line 157: Adding padding and avoiding magic numbers.
```suggestion
                    $output .= $this->border . str_repeat(' ', $this->padding);
                    $width = $width - strlen($this->border) - $this->padding;
```

src/Cli/TableFormatter.php
```diff
@@ -146,6 +178,10 @@ public function format($columns, $texts, $colors = [])
                 if ($column != $last) {
                     $output .= $this->border;
                 }
+
+                if ($this->isBorderedTable && $column === $last) {
+                    $output .= rtrim($this->border);
```
schlessera, line 183:
```suggestion
                    $output .= str_repeat(' ', $this->padding) . $this->border;
```

schlessera (changes requested):
This is all a bit hacky the way it is set up, but I think it is good enough for now in general. I'd just like to have the border and padding separated out into two separate properties and using this to get rid of magic numbers.

schlessera (approved):

PR #480, opened 2022-02-01
Allow Composer plugins to run

schlessera:
Composer now throws warnings when Composer plugins are being run, and will disable these at a later version for security reasons, unless they have been specifically allowed to run via the `composer.json` file.

This PR allows the plugins to run that we currently use as development dependencies:
- `civicrm/composer-downloads-plugin`
- `dealerdirect/phpcodesniffer-composer-installer`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,11 @@
     "mck89/peast": "Needed to minify the AMP script."
   },
   "config": {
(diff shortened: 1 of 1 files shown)


ediamin (approved):
LGTM 👍

PR #481, opened 2022-02-01
Remove test ignores that are not needed anymore

schlessera:
I noticed that some of the skipped tests were not broken anymore. This PR removes the ignores that are not needed anymore.


ediamin (approved):
Tested on my local. LGTM.

PR #504, opened 2022-02-25
Remove the override iterationArray property from the inherited classes

### LINKED ISSUE #503: Error when using AutoExtensions transformer
SS 2022-02-25 19 45 40

```
 $configurationData = [
        Configuration::KEY_TRANSFORMERS => [
          AutoExtensions::class,
          ServerSideRendering::class,
          TransformedIdentifier::class,
        ],
      ];
```

When I added AutoExtensions to Transformers, got an above error.
Then I fixed dupilicate of `$iterationArray` in `AmpProject/Validator/Spec/Section/Tags` and it is working well.

So, could you fix this problem?

ediamin:
Fixes #503 

Currently in PHP 5.6 we get errors in some validator spec classes due to override the property `$iterationArray` used in the trait `AmpProject\Validator\Spec\Iteration`.  This PR fixes this issue by removing the override property from the inherited classes.


ediamin:
In addition to the fix the error related to `iterationArray`, I've added the missing constants that are generated by the spec generator and also fix some PHP 8.1 compatibility issues.

src/Html/Attribute.php
```diff
@@ -74,7 +74,7 @@ interface Attribute
     const AMP_ONERROR                     = 'amp-onerror';
     const AMP_RUNTIME                     = 'amp-runtime';
     const AMP_SCRIPT_SRC                  = 'amp-script-src';
-    const AMP_STORY_DVH_POLLYFILL         = 'amp-story-dvh-polyfill';
+    const AMP_STORY_DVH_POLYFILL          = 'amp-story-dvh-polyfill';
```
schlessera, line 77: Good catch!

schlessera:
@ediamin Do you know why this was not caught by our PHP 5.6 tests? Can you try to add a test that triggers this issue on PHP 5.6 without the fix?

ediamin:
@schlessera I've found out that the `shivammathur/setup-php@v2` action uses `error_reporting = E_ALL & ~E_DEPRECATED & ~E_STRICT` and more specifically by ignoring the E_STRICT errors we don't see the issue with the `$iterationArray`. There is a [reported issue](https://github.com/shivammathur/setup-php/issues/450) in this action repo already. Further I've done testing using `error_reporting=E_ALL` in the test.yml. With this settings the tests for PHP 5.6 do not pass.

Should I update the `ini-values: pcov.directory=., error_reporting=E_ALL` for `shivammathur/setup-php@v2` action in tests.yml?

schlessera:
Yes, please. We want to catch all possible issues in our automated test suite.

ediamin:
Updated the `tests.yml` with `ini-values: pcov.directory=., error_reporting=E_ALL`

schlessera (approved):

src/Html/Attribute.php
westonruter: 🦜 wants a cracker

PR #523, opened 2022-05-12
Implement SSR for amp-audio

### LINKED ISSUE #518: Implement SSR for `amp-audio`
See:

* https://github.com/ampproject/amphtml/issues/29489
* https://github.com/ampproject/amphtml/pull/37982

ediamin:
Fixes #518 

This PR add the SSR support for amp-audio extension. It simply adds `<audio controls style="width: 100%"></audio>` element if amp-audio extension is present.

### DIFF
--- a/src/Optimizer/Transformer/ServerSideRendering.php
+++ b/src/Optimizer/Transformer/ServerSideRendering.php
@@ -142,6 +142,7 @@ final class ServerSideRendering implements Transformer
             if ($ampElement->tagName === Extension::AUDIO) {
                 $errors->add(Error\CannotRemoveBoilerplate::fromAmpAudio($ampElement));
                 $canRemoveBoilerplate = false;
+                $this->ssrAmpAudio($document, $ampElement);
             }
 
             /*
@@ -1059,5 +1060,29 @@ final class ServerSideRendering implements Transformer
         }
 
         return abs($number) < self::FLOATING_POINT_EPSILON;
+    }
+
+    /**
+     * Server-side rendering of an amp-audio element.
+     *
+     * @param Document $document DOM document to apply the transformations to.
+     * @param Element  $element  Element to adapt.
+     */
+    private function ssrAmpAudio(Document $document, Element $element)
+    {
+        // Check if we already have a SSR-ed audio element.
+        if ($element->hasChildNodes()) {
+            foreach ($element->childNodes as $childNode) {
+                if ($childNode instanceof Element && $childNode->tagName === Tag::AUDIO) {
+                    return;
+                }
+            }
+        }
+
+        $audio = $document->createElement(Tag::AUDIO);
+        $audio->setAttribute(Attribute::CONTROLS, '');
+        $audio->addInlineStyle('width:100%');
+
+        $element->appendChild($audio);
     }
 }
--- a/src/Optimizer/Transformer/ServerSideRendering.php
+++ b/src/Optimizer/Transformer/ServerSideRendering.php
@@ -1080,7 +1080,7 @@ final class ServerSideRendering implements Transformer
         }
 
         $audio = $document->createElement(Tag::AUDIO);
-        $audio->setAttribute(Attribute::CONTROLS, '');
+        $audio->setAttribute($document->createAttributeNode(Attribute::CONTROLS));
         $audio->addInlineStyle('width:100%');
 
         $element->appendChild($audio);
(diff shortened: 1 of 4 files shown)


ediamin:
@westonruter I would like to know your opinion about the boilerplate error. Should I remove the error for amp-audio or keep it as it is?

westonruter:
> I would like to know your opinion about the boilerplate error. Should I remove the error for amp-audio or keep it as it is?

I'm not sure I understand. As I understand, there still should be an error raised for `amp-audio` if it lacks a child `<audio controls style="width:100%;"></audio>`, correct?

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -1060,4 +1061,28 @@ private function isZero($number)
...
+        if ($element->hasChildNodes()) {
+            foreach ($element->childNodes as $childNode) {
+                if ($childNode instanceof Element && $childNode->tagName === Tag::AUDIO) {
+                    return;
+                }
+            }
+        }
+
+        $audio = $document->createElement(Tag::AUDIO);
+        $audio->setAttribute(Attribute::CONTROLS, '');
```
westonruter, line 1083: This should generate a boolean attribute rather than `controls=""`:

```suggestion
        $audio->setAttribute($document->createAttributeNode(Attribute::CONTROLS));
```

westonruter:
Oh I see. You're adding it if it isn't present originally. In that case, no, the error probably isn't necessary.

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -1060,4 +1061,28 @@ private function isZero($number)
...
+        $audio->addInlineStyle('width:100%');
```
westonruter, line 1084: Is the `style` necessary? I don't see it in https://github.com/ampproject/amphtml/pull/37982.

ediamin:
Sorry for the confusion. I would like to know your opinion about [these two lines](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/ServerSideRendering.php#L143-L144). Should I remove them or keep them as it?
```php
$errors->add(Error\CannotRemoveBoilerplate::fromAmpAudio($ampElement));
$canRemoveBoilerplate = false;
```

src/Optimizer/Transformer/ServerSideRendering.php
ediamin: Removed the inline style.

westonruter:
> I would like to know your opinion about [these two lines](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/ServerSideRendering.php#L143-L144). Should I remove them or keep them as it?

It seems to me that they can be removed, no? I don't see any reason why the error would be needed if we can always SSR the `amp-audio`.

ediamin:
> It seems to me that they can be removed, no? I don't see any reason why the error would be needed if we can always SSR the amp-audio.

I'm not sure about that. The comment says, `amp-audio requires knowing the dimensions of the browser. Do not remove the boilerplate or apply layout` and it is the same in the [amp-toolbox node repo](https://github.com/ampproject/amp-toolbox/blob/main/packages/optimizer/lib/transformers/ServerSideRendering.js#L84).

westonruter:
What I mean is that if amp-toolbox-php is _always_ adding the `audio` child to enable SSR, then there would never be a scenario where the presence of `amp-audio` would cause any error for SSR. So would the error message ever be correct?

ediamin:
I've updated the code and removed the error for amp-audio.

src/Optimizer/Transformer/ServerSideRendering.php
```diff
@@ -1060,4 +1058,28 @@ private function isZero($number)
...
+                if ($childNode instanceof Element && $childNode->tagName === Tag::AUDIO) {
+                    return;
+                }
+            }
+        }
+
+        $audio    = $document->createElement(Tag::AUDIO);
+        $controls = $document->createAttribute(Attribute::CONTROLS);
+
+        $audio->appendChild($controls);
```
westonruter, line 1082: Oh, I would have thought this would be:

```suggestion
        $audio->setAttributeNode($controls);
```

But it seems in this case `appendChild` works just the same: https://3v4l.org/kMYL4

westonruter (approved):
LGTM!

PR #524, opened 2022-05-17
Fix error when source is missing closing head tag

### LINKED ISSUE #522: Uncaught `DomException` in `Document::moveInvalidHeadNodesToBody`
When trying to parse `https://chuseok.org/places-to-visit-during-chuseok-mid-autumn-summer/` with `Document::fromHtml()`, I get an uncaught `DomException`:

```
PHP Fatal error:  Uncaught DOMException: Not Found Error in /vendor/ampproject/amp-toolbox/src/Dom/Document.php:553
```

This is the offending code:

https://github.com/ampproject/amp-toolbox-php/blob/89cd1f6bad3512fb8a08aae5fb6aedc64e988392/src/Dom/Document.php#L655-L669

PHPStorm warns about accessing `$node` potentially causing a null pointer exception too.

This fixes the bug for me:

```dif
-while ($node) {
+while ($node !== null) {
```

ediamin:
Fixes #522 

This PR fixes the uncaught `DOMException: Not Found Error` when the Document class tries to load a source that has no closing head and body tags and contains invalid node in the head tag. For example,
```html
<html><head><meta name="foo" content="bar"><foo></foo><bar></bar>
```
With this fix, we will get a proper structured html like this,
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <meta name="foo" content="bar" />
  </head>
  <body>
    <foo></foo><bar></bar>
  </body>
</html>
```

### DIFF
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -566,7 +566,13 @@ final class Document extends DOMDocument
         } elseif (! preg_match(self::HTML_STRUCTURE_BODY_END_TAG, $content, $matches)) {
             // Only <body> missing.
             // @todo This is an expensive regex operation, look into further optimization.
-            $content = preg_replace(self::HTML_STRUCTURE_HEAD_TAG, '$0<body>', $content, 1);
+            $content = preg_replace(self::HTML_STRUCTURE_HEAD_TAG, '$0<body>', $content, 1, $count);
+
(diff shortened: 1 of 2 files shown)


westonruter (approved):
LGTM

PR #531, opened 2022-08-15
Cache only valid responses

### LINKED ISSUE #530: i-amphtml-version=0 is invalid
If the metadata endpoint does not respond successfully, the current version ist set to '0':
https://github.com/ampproject/amp-toolbox-php/blame/main/src/RuntimeVersion.php#L80-L82

This is used by  the [AmpRuntimeCss.php](https://github.com/ampproject/amp-toolbox-php/blob/main/src/Optimizer/Transformer/AmpRuntimeCss.php#L162).

Atm I cannot provide a reproduction, but this occurs from time to time.

Desired behaviour:
Instead of setting the version to '0' it would be nice to throw an error/exception.

dritter:
Hi!

Recently we discovered that some requests to fetch the metadata failed, but were cached regardless of the status code. This PR adds a guard, that only valid responses are cached.

This fixes #530

### DIFF
--- a/src/RemoteRequest/CurlRemoteGetRequest.php
+++ b/src/RemoteRequest/CurlRemoteGetRequest.php
@@ -132,7 +132,7 @@ final class CurlRemoteGetRequest implements RemoteGetRequest
             $curlErrno = curl_errno($curlHandle);
             curl_close($curlHandle);
 
-            if ($body === false) {
+            if ($body === false || $status < 200 || $status >= 300) {
                 if (! $retriesLeft || in_array($curlErrno, self::RETRYABLE_ERROR_CODES, true) === false) {
                     if (! empty($status) && is_numeric($status)) {
                         throw FailedToGetFromRemoteUrl::withHttpStatus($url, (int) $status);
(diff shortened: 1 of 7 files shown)


dritter:
@schlessera Any chance to get this merged? This is quite urgent for us, because all our amp pages are invalid without this fix (see #530 ).

swissspidy:
cc @westonruter

src/RemoteRequest/CurlRemoteGetRequest.php
```diff
@@ -132,7 +132,7 @@ static function ($curl, $header) use (&$headers) {
             $curlErrno = curl_errno($curlHandle);
             curl_close($curlHandle);
 
-            if ($body === false) {
+            if ($body === false || $status < 200 || $status >= 300) {
```
westonruter, line 135: This should work as expected, but note that I believe `$status` could be `0`. In this case, `$body` should also be `false`.

westonruter (approved):
LGTM

src/RemoteRequest/CurlRemoteGetRequest.php
dritter: Hey! I am not sure, if I understand you correctly. Is there anything to do left?

If `$status` can be `0`, it should IMHO be treated as error. And this code does that (smaller than `200`).

westonruter: Yeah, I'm saying the same thing. In the error case, `$status` should be `0` and it will work as expected.

PR #533, opened 2022-08-24
PHP 8.2 Improvements

### LINKED ISSUE #532: PHP 8.2 Compatibility
PHP 8.2 is expected to be released at the end of November 2022.

Figured it can't hurt to track PHP 8.2 compatibility in a dedicated ticket.

At first glance it looks like the dynamic properties deprecation creates some warnings.

```
Creation of dynamic property AmpProject\Dom\Document::$head is deprecated in /path/to/vendor/ampproject/amp-toolbox/src/Dom/Document/Filter/DeduplicateTag.php on line 65

Creation of dynamic property AmpProject\Dom\Document::$body is deprecated in /path/to/vendor/ampproject/amp-toolbox/src/Dom/Document/Filter/DeduplicateTag.php on line 65

Creation of dynamic property AmpProject\Dom\Document::$html is deprecated in /path/to/vendor/ampproject/amp-toolbox/src/Dom/Document.php on line 699

Creation of dynamic property AmpProject\Dom\Document::$xpath is deprecated in /path/to/vendor/ampproject/amp-toolbox/src/Dom/Document.php on line 687
```

swissspidy:
Fixes #532

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
   "description": "A collection of AMP tools making it easier to publish and host AMP pages with PHP.",
   "license": "Apache-2.0",
   "require": {
-    "php": "^5.6 || ^7.0 || ^8.0 || ^8.1",
+    "php": "^5.6 || ^7.0 || ^8.0",
     "ext-dom": "*",
     "ext-filter": "*",
     "ext-iconv": "*",
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
   "description": "A collection of AMP tools making it easier to publish and host AMP pages with PHP.",
   "license": "Apache-2.0",
   "require": {
-    "php": "^5.6 || ^7.0 || ^8.0",
+    "php": "^5.6 || ^7.0 || ^8.0 || ^8.1",
     "ext-dom": "*",
     "ext-filter": "*",
     "ext-iconv": "*",
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -848,9 +848,6 @@ final class Document extends DOMDocument
             'ampCustomStyleByteCount',
             'inlineStyleByteCount',
             'links',
-            // Sometimes accessed by \Optimizer\Transformer\MinifyHtml::minifyNode().
-            // TODO: Revisit this.
-            'tagName',
         ];
     }
 
--- a/src/Dom/Document.php
+++ b/src/Dom/Document.php
@@ -110,7 +110,8 @@ final class Document extends DOMDocument
 
     /**
      * Associative array for lazily-created, cached properties for the document.
-     * @var
+     *
+     * @var array
      */
     private $properties = [];
 
@@ -641,22 +642,22 @@ final class Document extends DOMDocument
             }
 
             $this->body = $body;
-            $html->appendChild($this->body);
+            $html->appendChild($body);
 
-            if ($oldDocumentElement !== $this->body && $oldDocumentElement !== $this->head) {
-                $this->body->appendChild($oldDocumentElement);
+            if ($oldDocumentElement !== $body && $oldDocumentElement !== $this->head) {
+                $body->appendChild($oldDocumentElement);
             }
         } else {
             $head = $this->getElementsByTagName(Tag::HEAD)->item(0);
             if (!$head) {
-                $this->head = $this->createElement(Tag::HEAD);
-                $this->documentElement->insertBefore($this->head, $this->documentElement->firstChild);
+                $this->properties['head'] = $this->createElement(Tag::HEAD);
+                $this->documentElement->insertBefore($this->properties['head'], $this->documentElement->firstChild);
             }
 
             $body = $this->getElementsByTagName(Tag::BODY)->item(0);
             if (!$body) {
-                $this->body = $this->createElement(Tag::BODY);
-                $this->documentElement->appendChild($this->body);
+                $this->properties['body'] = $this->createElement(Tag::BODY);
+                $this->documentElement->appendChild($this->properties['body']);
             }
         }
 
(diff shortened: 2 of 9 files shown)


composer.json
```diff
@@ -4,7 +4,7 @@
   "description": "A collection of AMP tools making it easier to publish and host AMP pages with PHP.",
   "license": "Apache-2.0",
   "require": {
-    "php": "^5.6 || ^7.0 || ^8.0",
+    "php": "^5.6 || ^7.0 || ^8.0 || ^8.1",
```
swissspidy, line 7:
```suggestion
    "php": "^5.6 || ^7.0 || ^8.0",
```

src/Dom/Document.php
```diff
@@ -635,22 +642,22 @@ public function normalizeDomStructure()
...
-                $this->body->appendChild($oldDocumentElement);
+            if ($oldDocumentElement !== $body && $oldDocumentElement !== $this->head) {
+                $body->appendChild($oldDocumentElement);
             }
         } else {
             $head = $this->getElementsByTagName(Tag::HEAD)->item(0);
             if (!$head) {
-                $this->head = $this->createElement(Tag::HEAD);
-                $this->documentElement->insertBefore($this->head, $this->documentElement->firstChild);
+                $this->properties['head'] = $this->createElement(Tag::HEAD);
```
westonruter, line 653: Why not continue to use `$this->head` and `$this->body`? Is there a reason to switch to using `$this->properties`?

westonruter: Just for the sake of performance?

swissspidy: I ran into some issues when I originally tried that, but I might have been doing something wrong. I can give it another try 👍

src/Dom/Document.php
```diff
@@ -818,6 +828,32 @@ public function getRemainingCustomCssSpace()
...
+            Attribute::CHARSET,
+            Attribute::VIEWPORT,
+            'ampElements',
+            'ampCustomStyle',
+            'ampCustomStyleByteCount',
+            'inlineStyleByteCount',
+            'links',
+            // Sometimes accessed by \Optimizer\Transformer\MinifyHtml::minifyNode().
+            // TODO: Revisit this.
+            'tagName',
```
westonruter, line 853: Interesting. This probably indicates a bug in `minifyNode`. Namely, it should be checking if the node is an instance of `Element` before attempting to access `tagName`.

src/Dom/Document.php
swissspidy: Just tested again. Even on PHP 8.1 I am getting lots of `Undefined property` warnings in this case.

Even when doing something seemingly simple like this:

```php
// Throws warning
$this->html = $html;
return $this->html;

// Works
$this->html = $html;
return $html;

// Also works
$this->properties['html'] = $html;
return $this->properties['html'];
```

I also just analyzed the `AmpProject\Dom\DocumentTest::testNormalizeDomStructure()` test case for this.

1. Calling `normalizeDomStructure()` calls `$this->head`
2. This triggers `__get()`
3. The getter for `head` runs, and it sees that it isn't set yet, which causes a call to `normalizeDomStructure()` _again_. The property hasn't been set yet.
4. That one then calls `moveInvalidHeadNodesToBody` at the end
5. That one then triggers `Undefined property: AmpProject\Dom\Document::$head`.

No such issues happen with the current `$properties` approach.

src/Dom/Document.php
```diff
@@ -635,22 +642,22 @@ public function normalizeDomStructure()
             }
 
             $this->body = $body;
```
spacedmonkey, line 644:
```suggestion
            $this->properties['body'] = $body;
```
?

src/Dom/Document.php
```diff
@@ -108,6 +108,13 @@ final class Document extends DOMDocument
      */
     const XPATH_INLINE_STYLE_ATTRIBUTES_QUERY = './/@style';
 
+    /**
+     * Associative array for lazily-created, cached properties for the document.
+     *
+     * @var array
+     */
+    private $properties = [];
+
```
westonruter, line 117: I just realized a potential large simplification here. Why if each of the properties were instead just declared as null?

```suggestion
    public $xpath
    public $html;
    public $head;
    public $body;
    public $charset;
    public $viewport;
    public $ampElements;
    public $ampCustomStyle;
    public $ampCustomStyleByteCount;
    public $inlineStyleByteCount
    public $links;
```

The phpdoc not withstanding.

westonruter: I see you did this in 525e91853ce74d56cdd656ee08dd5494ed3ad200 but then reverted. Did you try `private` instead of `public`? This would model what is done with the `$links` property already.

westonruter: I see the problem with that approach. It's that private access in the class won't cause the getter to be invoked. I think I have a way around that, and that is to manually invoke `__get()` for methods on `Document`. I'm going to give it a try, though the commit may not be pushed until Monday.

westonruter: Bah. Doing this makes PHPStan blow up with a lot of "Access to private property" errors.

src/Dom/Document.php
westonruter: That will bypass the `__set()` method, but above this method is already doing `$this->head`. Since elsewhere in this method the magic getters/setters are being bypassed, probably good to do so here too.

src/Dom/Document.php
westonruter: I think the issue at question is this [note](https://www.php.net/manual/en/language.oop5.overloading.php#object.get:~:text=PHP%20will%20not%20call%20an%20overloaded%20method%20from%20within%20the%20same%20overloaded%20method.):

> PHP will not call an overloaded method from within the same overloaded method. That means, for example, writing return $this->foo inside of [__get()](https://www.php.net/manual/en/language.oop5.overloading.php#object.get) will return null and raise an E_WARNING if there is no foo property defined, rather than calling [__get()](https://www.php.net/manual/en/language.oop5.overloading.php#object.get) a second time. However, overload methods may invoke other overload methods implicitly (such as [__set()](https://www.php.net/manual/en/language.oop5.overloading.php#object.set) triggering [__get()](https://www.php.net/manual/en/language.oop5.overloading.php#object.get)).

westonruter: In other words, if a caller is trying to access `$x->foo` which invokes a magic getter, the getter must not also try to access `$x->foo` although it can access `$x->bar`. So there can't be any recursive reference to the current magic property being gotten.

westonruter (approved):

PR #536, opened 2022-10-19
Fix PHPStan errors

thelovekesh:
This PR aims to fix PHPStan errors introduced as a part of https://github.com/ampproject/amp-toolbox-php/pull/534. See the [`error log`](https://github.com/ampproject/amp-toolbox-php/actions/runs/3277984591/jobs/5395871464#step:7:14).

image

## Tasks Done
- Add a guard to check if passed `$url` in `CurlRemoteGetRequest::get` is a `non-empty-string`.
- Remove `curl_setopt($curlHandle, CURLOPT_HEADER, 0);` as we need to pass `true` if we want to enable this option. ([Ref](https://www.php.net/manual/en/function.curl-setopt.php#:~:text=CURLOPT_HEADER,in%20the%20output.)).
- Update the `CURLOPT_SSL_VERIFYPEER` value as we need to pass `false` if we want to disable it. ([Ref](https://www.php.net/manual/en/function.curl-setopt.php#:~:text=CURLOPT_SSL_VERIFYPEER,the%20peer%27s%20certificate.)).

### DIFF
--- a/src/RemoteRequest/CurlRemoteGetRequest.php
+++ b/src/RemoteRequest/CurlRemoteGetRequest.php
@@ -108,7 +108,7 @@ final class CurlRemoteGetRequest implements RemoteGetRequest
 
             curl_setopt($curlHandle, CURLOPT_URL, $url);
             curl_setopt($curlHandle, CURLOPT_FOLLOWLOCATION, true);
-            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify ? true : false);
+            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify);
             curl_setopt($curlHandle, CURLOPT_SSL_VERIFYHOST, $this->sslVerify ? 2 : 0);
             curl_setopt($curlHandle, CURLOPT_RETURNTRANSFER, true);
             curl_setopt($curlHandle, CURLOPT_FAILONERROR, true);
--- a/src/RemoteRequest/CurlRemoteGetRequest.php
+++ b/src/RemoteRequest/CurlRemoteGetRequest.php
@@ -108,7 +108,7 @@ final class CurlRemoteGetRequest implements RemoteGetRequest
 
             curl_setopt($curlHandle, CURLOPT_URL, $url);
             curl_setopt($curlHandle, CURLOPT_FOLLOWLOCATION, true);
-            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify ? 1 : false);
+            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify ? true : false);
             curl_setopt($curlHandle, CURLOPT_SSL_VERIFYHOST, $this->sslVerify ? 2 : 0);
             curl_setopt($curlHandle, CURLOPT_RETURNTRANSFER, true);
             curl_setopt($curlHandle, CURLOPT_FAILONERROR, true);
--- a/src/RemoteRequest/CurlRemoteGetRequest.php
+++ b/src/RemoteRequest/CurlRemoteGetRequest.php
@@ -108,7 +108,7 @@ final class CurlRemoteGetRequest implements RemoteGetRequest
 
             curl_setopt($curlHandle, CURLOPT_URL, $url);
             curl_setopt($curlHandle, CURLOPT_FOLLOWLOCATION, true);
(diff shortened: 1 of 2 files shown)


src/RemoteRequest/CurlRemoteGetRequest.php
```diff
@@ -97,13 +98,17 @@ public function __construct($sslVerify = true, $timeout = self::DEFAULT_TIMEOUT,
...
+        }
+
         do {
             $curlHandle = curl_init();
 
             curl_setopt($curlHandle, CURLOPT_URL, $url);
-            curl_setopt($curlHandle, CURLOPT_HEADER, 0);
             curl_setopt($curlHandle, CURLOPT_FOLLOWLOCATION, true);
-            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify ? 1 : 0);
+            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify ? 1 : false);
```
westonruter, line 111: And `true` shouldn't be passed instead of `1`?

thelovekesh: Hmm. I have updated it.

src/RemoteRequest/CurlRemoteGetRequest.php
```diff
@@ -97,13 +98,17 @@ public function __construct($sslVerify = true, $timeout = self::DEFAULT_TIMEOUT,
...
+        }
+
         do {
             $curlHandle = curl_init();
 
             curl_setopt($curlHandle, CURLOPT_URL, $url);
-            curl_setopt($curlHandle, CURLOPT_HEADER, 0);
             curl_setopt($curlHandle, CURLOPT_FOLLOWLOCATION, true);
-            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify ? 1 : 0);
+            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify ? true : false);
```
westonruter, line 111: Seems this could now just be simply:

```suggestion
            curl_setopt($curlHandle, CURLOPT_SSL_VERIFYPEER, $this->sslVerify);
```

src/RemoteRequest/CurlRemoteGetRequest.php
```diff
@@ -97,13 +98,17 @@ public function __construct($sslVerify = true, $timeout = self::DEFAULT_TIMEOUT,
...
+
+        if (! is_string($url) || empty($url)) {
+            throw FailedToGetFromRemoteUrl::withException($url, FailedToParseUrl::forUrl($url));
+        }
+
         do {
             $curlHandle = curl_init();
 
             curl_setopt($curlHandle, CURLOPT_URL, $url);
-            curl_setopt($curlHandle, CURLOPT_HEADER, 0);
```
westonruter, line 104: Would passing `false` here also work? I tried looking for any case where `CURLOPT_HEADER` would be enabled by default, but I couldn't. Still, it seems perhaps passing `false` would be a bit safer to be sure?

src/RemoteRequest/CurlRemoteGetRequest.php
```diff
@@ -97,13 +98,17 @@ public function __construct($sslVerify = true, $timeout = self::DEFAULT_TIMEOUT,
     public function get($url, $headers = [])
     {
         $retriesLeft = $this->retries;
+
+        if (! is_string($url) || empty($url)) {
+            throw FailedToGetFromRemoteUrl::withException($url, FailedToParseUrl::forUrl($url));
+        }
```
westonruter, line 104: This is good. 

Do you think it would be overkill to also check the `$url` with `parse_url()`? I suppose so since cURL itself would fail and return a better error code in that case?

src/RemoteRequest/CurlRemoteGetRequest.php
thelovekesh: It's false by default so doesn't really matter. It only shows headers when it's passed as `true`. But we can keep it with `false` for visibility.

image

src/RemoteRequest/CurlRemoteGetRequest.php
thelovekesh: We can add `parse_url()` if we want to avoid a curl request with the wrong URL and throw the error beforehand. 

> I suppose so since cURL itself would fail and return a better error code in that case?

cURL fails but the exception looks like this. I think we can add it with a more clear message like `Passed string is not a valid URL`.

image

westonruter: I think this is good. The error coming from cURL will be more descriptive.

westonruter (approved):
Thank you!

PR #544, opened 2023-10-20
Add `fetchpriority=high` on hero images

thelovekesh:
See https://github.com/ampproject/amp-wp/issues/7601

### DIFF
--- a/src/Optimizer/Transformer/OptimizeHeroImages.php
+++ b/src/Optimizer/Transformer/OptimizeHeroImages.php
@@ -114,6 +114,13 @@ final class OptimizeHeroImages implements Transformer
     const NOSCRIPT_IMG_XPATH_QUERY = './noscript/img';
 
     /**
+     * XPath query to relatively fetch all noscript > img[fetchpriority="high"] elements.
+     *
+     * @var string
+     */
+    const NOSCRIPT_IMG_FETCHPRIORITY_HIGH_XPATH_QUERY = './noscript/img[@fetchpriority="high"]';
+
+    /**
      * Regular expression pattern to extract the URL from a CSS background-image property.
      *
      * @var string
@@ -213,6 +220,17 @@ final class OptimizeHeroImages implements Transformer
 
             if ($node->tagName === Tag::P) {
                 $seenParagraphCount++;
+            }
+
+            if ($node->tagName === Extension::IMG && ! $node->hasAttribute(Attribute::DATA_HERO)) {
+                $highFetchpriorityImage = $document->xpath->query(
+                    self::NOSCRIPT_IMG_FETCHPRIORITY_HIGH_XPATH_QUERY,
+                    $node
+                )->item(0);
+
+                if ($highFetchpriorityImage instanceof Element) {
+                    $node->setAttribute(Attribute::DATA_HERO, '');
+                }
             }
 
             $heroImage = $this->detectImageWithAttribute($node, Attribute::DATA_HERO);
--- a/src/Optimizer/Transformer/OptimizeHeroImages.php
+++ b/src/Optimizer/Transformer/OptimizeHeroImages.php
@@ -229,7 +229,7 @@ final class OptimizeHeroImages implements Transformer
                 )->item(0);
 
                 if ($highFetchpriorityImage instanceof Element) {
-                    $node->setAttribute(Attribute::DATA_HERO, '');
+                    $node->appendChild($document->createAttribute(Attribute::DATA_HERO));
                 }
             }
 
--- a/src/Optimizer/Transformer/OptimizeHeroImages.php
+++ b/src/Optimizer/Transformer/OptimizeHeroImages.php
@@ -566,6 +566,14 @@ final class OptimizeHeroImages implements Transformer
                 $imgElement->setAttribute(Attribute::LOADING, $noscript_img->getAttribute(Attribute::LOADING));
             }
 
+            // Preserve the original fetchpriority attribute from the noscript fallback img.
+            if ($noscript_img->hasAttribute(Attribute::FETCHPRIORITY)) {
+                $imgElement->setAttribute(
+                    Attribute::FETCHPRIORITY,
+                    $noscript_img->getAttribute(Attribute::FETCHPRIORITY)
+                );
+            }
+
             // Remove any noscript>img when an amp-img is pre-rendered.
             $noscript_img->parentNode->parentNode->removeChild($noscript_img->parentNode);
         } elseif (! $this->isMarkedAsHeroImage($element)) {
--- a/tests/Optimizer/Transformer/OptimizeHeroImagesTest.php
+++ b/tests/Optimizer/Transformer/OptimizeHeroImagesTest.php
@@ -342,12 +342,25 @@ final class OptimizeHeroImagesTest extends TestCase
                 ),
             ],
 
-            'preserves fetchpriority attribute from noscript fallback img on prerendered image' => [
+            'preserves fetchpriority attribute from noscript fallback img' => [
                 $input(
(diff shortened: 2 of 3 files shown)


src/Optimizer/Transformer/OptimizeHeroImages.php
```diff
@@ -555,6 +555,7 @@ private function generateImg(HeroImage $heroImage, Document $document)
         $imgElement = $document->createElement(Tag::IMG);
         $imgElement->setAttribute(Attribute::CLASS_, self::SSR_IMAGE_CLASS);
         $imgElement->setAttribute(Attribute::DECODING, 'async');
+        $imgElement->setAttribute(Attribute::FETCHPRIORITY, 'high');
```
westonruter, line 558: Thinking about this some more now... perhaps we actually shouldn't automatically add `fetchpriority=high` to all hero images, since there can be multiple. The `fetchpriority=high` attribute should be reserved for the LCP image. So I think instead of automatically adding it to all heros, we should instead only it when the `amp-img > noscript > img` contains `fetchpriority=high`.

Additionally, we should automatically SSR any `amp-img` that has a `noscript > img` with `fetchpriority=high`.

But if an `amp-img` has `data-hero` alone, that wouldn't necessarily warrant adding `fetchpriority=high`.

thelovekesh: > Thinking about this some more now... perhaps we actually shouldn't automatically add fetchpriority=high to all hero images, since there can be multiple. 

Completely agree here and I have updated it per suggestion.

westonruter (approved):
Looks good!

src/Optimizer/Transformer/OptimizeHeroImages.php
```diff
@@ -215,6 +222,17 @@ private function findHeroImages(Document $document, $maxHeroImageCount)
...
+            if ($node->tagName === Extension::IMG && ! $node->hasAttribute(Attribute::DATA_HERO)) {
+                $highFetchpriorityImage = $document->xpath->query(
+                    self::NOSCRIPT_IMG_FETCHPRIORITY_HIGH_XPATH_QUERY,
+                    $node
+                )->item(0);
+
+                if ($highFetchpriorityImage instanceof Element) {
+                    $node->setAttribute(Attribute::DATA_HERO, '');
+                }
+            }
```
thelovekesh, line 234: With this logic in place, we won't need to add `data-hero` to images with high fetch priority in AMP Image sanitizer.

tests/Optimizer/Transformer/OptimizeHeroImagesTest.php
```diff
@@ -341,6 +341,28 @@ public function dataTransform()
...
+
+            'preserves fetchpriority attribute from noscript fallback img' => [
+                $input(
+                    '<amp-img width="500" height="400" src="/img1.png"><noscript><img src="/img1.png" width="500" height="400" fetchpriority="high"></noscript></amp-img>'
+                ),
+                $output(
+                    '<amp-img data-hero width="500" height="400" src="/img1.png" i-amphtml-ssr><img class="i-amphtml-fill-content i-amphtml-replaced-content" decoding="async" fetchpriority="high" src="/img1.png"></amp-img>'
+                ),
+            ],
+
```
westonruter, line 353: Am I right that this test is somewhat redundant with the second?

tests/Optimizer/Transformer/OptimizeHeroImagesTest.php
```diff
@@ -341,6 +341,28 @@ public function dataTransform()
...
+            'automatically ssr img element with high fetchpriority attribute from noscript fallback img' => [
+                $input(
+                    '<amp-img width="500" height="400" src="/img1.png"><noscript><img src="/img1.png" width="500" height="400" fetchpriority="high"></noscript></amp-img>'
+                    .  '<amp-img width="500" height="400" src="/img2.png"><noscript><img src="/img2.png" width="500" height="400"></noscript></amp-img>'
+                    .  '<amp-img width="500" height="400" src="/img3.png"><noscript><img src="/img3.png" width="500" height="400" fetchpriority="high"></noscript></amp-img>'
```
westonruter, line 358: For the second image, how about adding `data-hero` to it to see what happens when preceded by a `fetchpriority=high` image, and then the third image can have neither `data-hero` nor `fetchpriority` (as really the source markup shouldn't have multiple `fetchpriority=high`.

src/Optimizer/Transformer/OptimizeHeroImages.php
```diff
@@ -215,6 +222,17 @@ private function findHeroImages(Document $document, $maxHeroImageCount)
...
+
             $heroImage = $this->detectImageWithAttribute($node, Attribute::DATA_HERO);
             if ($heroImage) {
```
westonruter, line 237: Alternatively, this could be done without setting `data-hero`, right?

```suggestion

            $hasFetchpriorityHigh = false;
            if ($node->tagName === Extension::IMG && ! $node->hasAttribute(Attribute::DATA_HERO)) {
                $highFetchpriorityImage = $document->xpath->query(
                    self::NOSCRIPT_IMG_FETCHPRIORITY_HIGH_XPATH_QUERY,
                    $node
                )->item(0);

                if ($highFetchpriorityImage instanceof Element) {
                    $hasFetchpriorityHigh = true;
                }
            }

            $heroImage = $this->detectImageWithAttribute($node, Attribute::DATA_HERO);
            if ($heroImage || $hasFetchpriorityHigh) {
```

If absent, then `generateImg` will be adding it anyway:

https://github.com/ampproject/amp-toolbox-php/blob/89707d0e55a4cbb4be6c1b3a3ed23d9193beb4a6/src/Optimizer/Transformer/OptimizeHeroImages.php#L601-L603

thelovekesh: But here the node should be transformed to HeroImage. 

https://github.com/ampproject/amp-toolbox-php/blob/89707d0e55a4cbb4be6c1b3a3ed23d9193beb4a6/src/Optimizer/Transformer/OptimizeHeroImages.php#L288

Adding `data-hero` will do that.

tests/Optimizer/Transformer/OptimizeHeroImagesTest.php
thelovekesh: Yes.

src/Optimizer/Transformer/OptimizeHeroImages.php
```diff
@@ -215,6 +222,17 @@ private function findHeroImages(Document $document, $maxHeroImageCount)
...
             }
 
+            if ($node->tagName === Extension::IMG && ! $node->hasAttribute(Attribute::DATA_HERO)) {
+                $highFetchpriorityImage = $document->xpath->query(
+                    self::NOSCRIPT_IMG_FETCHPRIORITY_HIGH_XPATH_QUERY,
+                    $node
+                )->item(0);
+
+                if ($highFetchpriorityImage instanceof Element) {
+                    $node->setAttribute(Attribute::DATA_HERO, '');
```
westonruter, line 232: This creates a boolean attribute without any value:

```suggestion
                    $node->appendChild($document->createAttribute(Attribute::DATA_HERO));
```

tests/Optimizer/Transformer/OptimizeHeroImagesTest.php
thelovekesh: Updated the test cases in https://github.com/ampproject/amp-toolbox-php/pull/544/commits/16b9f6689b66031fd09d512955fb1bae36236115

westonruter (approved):
🎉

